PR #52, opened 2025-05-13
Muscle3 actor

maarten-ic:
Implement a [muscle3](https://muscle3.readthedocs.io/en/latest/) actor for the waveform editor

### DIFF
--- a/docs/source/muscle3.rst
+++ b/docs/source/muscle3.rst
@@ -1,0 +1,77 @@
+MUSCLE3 IMAS Actor
+==================
+
+The waveform editor includes an actor that can be included in an IMAS `MUSCLE3
+<https://muscle3.readthedocs.io/>`__ simulation.
+
+.. caution::
+    The IMAS MUSCLE3 actor requires the following packages:
+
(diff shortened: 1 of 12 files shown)


mikesndrs (approved):

docs/source/muscle3.rst
```diff
@@ -0,0 +1,77 @@
...
+        executable: waveform-editor
+        args: actor
+
+Available settings
+''''''''''''''''''
+
+- ``waveforms`` (mandatory): indicate the (full) path to the waveform configuration.
+
+
+Input ports (``F_INIT``)
```
SBlokhuizen, line 39: It might be nice to add the MUSCLE3 block diagram from the slides of previous progress meeting to the docs.

SBlokhuizen (commented):
Looks good, one minor comment

docs/source/muscle3.rst
maarten-ic: As discussed, since there are no fixed input/output port names, a block diagram with example `ec_launchers_out` en `nbi_out` ports could be confusing for users (since any IDS can appear as output if there are waveforms defined for that IDS).

PR #55, opened 2025-05-14
Improve startup times of the CLI

maarten-ic:
Move slow GUI imports into the `gui` command to speed up the CLI for the other use cases.


SBlokhuizen (approved):
Nice! This is quite a bit faster now!

PR #60, opened 2025-06-03
Feature/piecewise drag and drop

SBlokhuizen:
This add drag & drop functionality to piecewise linear tendencies. If the waveform contains a piecewise linear tendency, and it is opened in the editor, the "Point Draw Tool" is made available in the plotting window. Multiple piecewise linear tendencies are supported. 

Some example cases:


MireilleSchneider (approved):

SBlokhuizen:
As discussed with @maarten-ic we will first do a major GUI refactor (already in process with #65 and #66) before we merge this functionality.

PR #68, opened 2025-06-06
Add click and drag for piecewise linear tendencies

SBlokhuizen:
- Added click & drag functionality to piecewise linear tendencies.
- Supports multiple piecewise linear tendencies.
- As `plot_waveform` is rewritten, there is little shared functionality for the `PlotterEdit` and `PlotterView`, so  the plotter base class was removed.


waveform_editor/gui/plotter_edit.py
```diff
@@ -1,16 +1,28 @@
...
+from panel.viewable import Viewer
+from ruamel.yaml import YAML
 
-from waveform_editor.gui.plotter_base import PlotterBase
+from waveform_editor.tendencies.piecewise import PiecewiseLinearTendency
 from waveform_editor.waveform import Waveform
 
 
-class PlotterEdit(PlotterBase):
+class PlotterEdit(Viewer):
```
maarten-ic, line 14: Drag and drop had quite some issues. As discussed, I've tried to fix them in #69

waveform_editor/gui/plotter_view.py
```diff
@@ -30,3 +31,26 @@ def update_plot(self):
...
+        ylabel = "Value"
+
+        if waveform is None or not waveform.tendencies:
+            return hv.Curve(([], []), xlabel, ylabel)
+        times, values = waveform.get_value()
+
+        return hv.Curve((times, values), xlabel, ylabel, label=waveform.name)
+
+    def __panel__(self):
+        return pn.Column(self.pane)
```
maarten-ic, line 56:
```suggestion
        return self.pane
```

maarten-ic: Okay, makes sense!  Can you add a short comment in the code to explain this? :)

PR #73, opened 2025-06-19
Docs/update installation

maarten-ic:
Updated the installation guide.

Note that I have already put in `pip install waveform-editor`, even though that does not work yet ;)

@MireilleSchneider I have added a note for SDCC install: `module load IMAS-Python` is enough to get it to work.


docs/source/installing.rst
```diff
@@ -9,28 +9,28 @@ This section guides you through installing the Waveform Editor package.
 Standard Installation
 ---------------------
 
-.. warning:: Installing from PyPI is not yet supported currently. Please follow `Installation from Source (for Development)`_ below.
+The recommended way to install the Waveform Editor is pip installing from from PyPI.
```
SBlokhuizen, line 12:
```suggestion
The recommended way to install the Waveform Editor is pip installing from PyPI.
```

SBlokhuizen (commented):
One typo, the rest looks good to me

PR #74, opened 2025-06-19
Fail CI if there are warnings when building docs

SBlokhuizen:


waveform_editor/util.py
```diff
@@ -43,17 +43,19 @@ class State:
...
-            if ignore_clicks:
-                return
-            ...
-        widget = pn.widgets.Button(name="Process", on_click=process)
-
-        # Elsewhere we can use a with statement to temporarily ignore clicks:
-        with ignore_clicks:
-            # process will not do anything, even though the on_click triggers
-            widget.clicks = 0
+        .. code-block:: bash
```
maarten-ic, line 46:
```suggestion
        .. code-block:: python
```

PR #77, opened 2025-06-30
Derived Waveforms

SBlokhuizen:
- Implementation of `DerivedWaveform` class, which inherits from the new `BaseWaveform` class
- Introduction of `DependencyGraph`, which is responsible for checking if there are any cycles for the derived waveforms graph
- Constants (i.e. waveforms from an int/float) are now treated as derived waveforms instead.

The following YAML shows some examples containing derived waveforms:
```yaml
globals:
  dd_version: 4.0.0
ec_launchers:
  beams:
    test/1:
    - {type: linear, from: 10, to: 50, duration: 20}
    - {type: constant, value: 50, duration: 20}
    - {type: linear, from: 50, to: 30, duration: 20}
    test/2:
    - {type: sine, amplitude: 3, duration: 60}
    test/3: |
      'test/1'
    test/4: |
      'test/5' / 2
    test/5: |
      np.maximum('test/1', 40)
    test/6: |
      2 * 'test/1' + 'test/2' / 2
    test/7: 2
```
Still to do:
- [x] Write docstrings
- [x] Write tests
- [x] Add section to documentation

### DIFF
--- a/waveform_editor/derived_waveform.py
+++ b/waveform_editor/derived_waveform.py
@@ -6,6 +6,12 @@ from asteval import Interpreter
 
 from waveform_editor.base_waveform import BaseWaveform
 
+NUMPY_UFUNCS = {}
+for name in np.__all__:
+    obj = getattr(np, name)
+    if isinstance(obj, np.ufunc):
+        NUMPY_UFUNCS[name] = obj
+
 
 class DependencyRenamer(ast.NodeTransformer):
     """AST transformer to rename string constants."""
@@ -67,9 +73,6 @@ class DerivedWaveform(BaseWaveform):
         self.dependencies = set()
         self.is_constant = False
         self.expression = None
-        self.user_symbols = {
-            name: obj for name, obj in vars(np).items() if isinstance(obj, np.ufunc)
-        }
         self.prepare_expression()
 
     def prepare_expression(self):
@@ -142,9 +145,10 @@ class DerivedWaveform(BaseWaveform):
             return time, np.zeros_like(time)
 
         eval_context = self._build_eval_context(time)
-        self.user_symbols["__w"] = eval_context
+        sym_table = NUMPY_UFUNCS.copy()
+        sym_table["__w"] = eval_context
         aeval = Interpreter(
-            symtable=self.user_symbols,
+            symtable=sym_table,
             minimal=True,
             use_numpy=False,
         )


waveform_editor/gui/plotter_edit.py
```diff
@@ -39,6 +42,13 @@ def update_plot(self):
         if self._update_plot_from_drag:
             return  # Skip update triggered from a drag-and-drop
 
+        if isinstance(self.plotted_waveform, DerivedWaveform):
+            try:
+                self.pane.object = self.main_curve()
+            except Exception as e:
+                self.editor.create_error_alert(e)
+            return
```
maarten-ic, line 50: Let's just install a generic error handler instead of these.
See https://stackoverflow.com/questions/57339953/raise-errors-and-display-traceback-messages-from-panel-dashboards#72167334

waveform_editor/gui/editor.py
```diff
@@ -77,6 +78,9 @@ def on_value_change(self, event):
         # dashed lists are placed below the key containing the waveform name
         if editor_text.lstrip().startswith("- "):
             waveform_yaml = f"{name}:\n{editor_text}"
+        # Derived waveforms are parsed as YAML block strings
+        elif "'" in editor_text or '"' in editor_text:
+            waveform_yaml = f"{name}: |\n  {editor_text}"
```
maarten-ic, line 83: What does this do if you have a regular waveform and accidentally add a ' or " (by typo)?

waveform_editor/gui/editor.py
```diff
@@ -85,22 +89,32 @@ def on_value_change(self, event):
...
+                    "There was an error in the YAML configuration\n" + f"{annotations}",
+                    "warning",
                 )
-                self.error_alert.alert_type = "warning"
             else:
                 self.error_alert.object = ""  # Clear any previous errors or warnings
             # There are no errors: update self.waveform
             self.waveform = waveform
 
+    def create_error_alert(self, message, alert_type="danger"):
```
maarten-ic, line 108: Let's remove the default argument?
```suggestion
    def create_error_alert(self, message, alert_type):
```

waveform_editor/gui/plotter_view.py
```diff
@@ -46,7 +48,9 @@ def plot_waveform(self, waveform):
         xlabel = "Time (s)"
         ylabel = "Value"
 
-        if waveform is None or not waveform.tendencies:
+        if waveform is None or (
+            isinstance(waveform, Waveform) and not waveform.tendencies
+        ):
             return hv.Curve(([], []), xlabel, ylabel)
         times, values = waveform.get_value()
```
maarten-ic, line 55: Perhaps we could instead update the get_value() logic in the Waveform class to return `([], [])` when it has no tendencies and remove the isinstance and waveform.tendencies check here?

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self):
+        self.graph = {}
+
+    def add_node(self, name, dependencies):
+        self.graph[name] = set(dependencies)
+        try:
+            self.detect_cycles()
+        except RuntimeError as e:
+            del self.graph[name]
+            raise e
```
maarten-ic, line 14:
```suggestion
        except RuntimeError:
            del self.graph[name]
            raise
```

Otherwise the stacktrace is a bit weird:
```python
>>> try:
...     raise RuntimeError("test")
... except RuntimeError:
...     raise
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
RuntimeError: test
>>> try:
...     raise RuntimeError("test")
... except RuntimeError as e:
...     raise e
... 
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
  File "<stdin>", line 2, in <module>
RuntimeError: test
```

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def remove_node(self, name):
+        if name in self.graph:
```
maarten-ic, line 17: Wouldn't it be an error if `name` doesn't exist in self.graph?

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+            del self.graph[name]
+        for deps in self.graph.values():
+            deps.discard(name)
```
maarten-ic, line 20: We shouldn't do this, right?
The underlying waveform still has a dependency on `name`.

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def rename_node(self, old_name, new_name, config):
+        dependents = {node for node, deps in self.graph.items() if old_name in deps}
+        for dependent_name in dependents:
+            dependent_waveform = config[dependent_name]
+            if isinstance(dependent_waveform, DerivedWaveform):
+                dependent_waveform.rename_dependency(old_name, new_name)
```
maarten-ic, line 27: This makes it spaghetti, let's just return the dependents list and execute this logic in the Configuration

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+            raise e
+
+    def remove_node(self, name):
+        if name in self.graph:
+            del self.graph[name]
+        for deps in self.graph.values():
+            deps.discard(name)
+
+    def rename_node(self, old_name, new_name, config):
+        dependents = {node for node, deps in self.graph.items() if old_name in deps}
```
maarten-ic, line 23:
```suggestion
        dependents = [node for node, deps in self.graph.items() if old_name in deps]
```

Lists are more efficient to construct and loop over than sets.

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+        for dependent_name in dependents:
+            dependent_waveform = config[dependent_name]
+            if isinstance(dependent_waveform, DerivedWaveform):
+                dependent_waveform.rename_dependency(old_name, new_name)
+
+        if old_name in self.graph:
+            self.graph[new_name] = self.graph.pop(old_name)
+
+        for dependencies in self.graph.values():
+            if old_name in dependencies:
```
maarten-ic, line 33: You already determined this and stored it in the dependents list. Please reuse it :)

waveform_editor/dependency_graph.py
```diff
@@ -0,0 +1,65 @@
...
+        else:
+            for node in self.graph:
+                visit(node)
+
+    def print(self):
+        for node, deps in self.graph.items():
+            if deps:
+                print(f"{node} -> {', '.join(deps)}")
+            else:
+                print(f"{node} -> (no dependencies)")
```
maarten-ic, line 65: You could just use `pprint.pprint(self.graph)`?

waveform_editor/configuration.py
```diff
@@ -216,6 +293,24 @@ def _to_commented_map(self):
...
+        for name in self.waveform_map:
+            waveform = self[name]
+            if not isinstance(waveform, DerivedWaveform) and waveform.tendencies:
+                start = waveform.tendencies[0].start
+                end = waveform.tendencies[-1].end
+
+                if start < min_start:
+                    min_start = start
+                if end > max_end:
+                    max_end = end
```
maarten-ic, line 309: You can make this a bit shorter
```suggestion
                min_start = min(min_start, waveform.tendencies[0].start)
                max_end = max(max_end, waveform.tendencies[-1].end)
```

waveform_editor/configuration.py
```diff
@@ -69,6 +79,16 @@ def add_waveform(self, waveform, path):
         group = self.traverse(path)
         group.waveforms[waveform.name] = waveform
         self.waveform_map[waveform.name] = group
+        if isinstance(waveform, DerivedWaveform):
+            try:
+                self.dependency_graph.add_node(
+                    waveform.name, waveform.dependent_waveforms
+                )
```
maarten-ic, line 86: Can you do this check before adding it to `group.waveforms` and `self.waveform_map`? Then you can remove the try/except as you wouldn't need to execute the `del` statements.

waveform_editor/configuration.py
```diff
@@ -115,9 +138,24 @@ def replace_waveform(self, waveform):
...
+            if isinstance(waveform, DerivedWaveform):
+                self.dependency_graph.add_node(
+                    waveform.name, waveform.dependent_waveforms
+                )
+            else:
+                self.dependency_graph.remove_node(waveform.name)
+            self._calculate_bounds()
+        except Exception as e:
+            # Revert replacement
+            group.waveforms[waveform.name] = old_waveform
```
maarten-ic, line 157: This risks desynchronizing your dependency_graph (which will have removed `waveform.name` when add_node failed).

waveform_editor/configuration.py
```diff
@@ -115,9 +138,24 @@ def replace_waveform(self, waveform):
...
-            group = self.waveform_map[waveform.name]
-            group.waveforms[waveform.name] = waveform
+
+        group = self.waveform_map[waveform.name]
+        old_waveform = group.waveforms[waveform.name]
+
+        group.waveforms[waveform.name] = waveform
+
+        try:
+            if isinstance(waveform, DerivedWaveform):
```
maarten-ic, line 148: Can we check in the editor whether the DerivedWaveform that you're editing is circular and avoid calling this method in that case altogether? That would make this method a lot easier and avoid al this error handling and reverting?

waveform_editor/configuration.py
```diff
@@ -128,9 +166,20 @@ def remove_waveform(self, name):
         if name not in self.waveform_map:
             raise ValueError(f"Waveform '{name}' does not exist in the configuration.")
 
+        # Check if any DerivedWaveform depends on this waveform
+        for group in self.waveform_map.values():
+            for wf in group.waveforms.values():
```
maarten-ic, line 171: Now you're doing too much work (which can scale quadratically).

```python
for name, group in self.waveform_map.items():
    wf = group[name]
    ...
```

maarten-ic: Wouldn't it be easier to query the dependency_graph instead?

waveform_editor/configuration.py
```diff
@@ -139,8 +188,36 @@ def remove_group(self, path):
...
         """
         parent_group = self if len(path) == 1 else self.traverse(path[:-1])
-        group = parent_group.groups.pop(path[-1])
+        group = parent_group.groups[path[-1]]
+
+        to_remove = self._collect_waveforms_in_group(group)
+
+        for wf_name, grp in self.waveform_map.items():
+            if wf_name not in to_remove:
+                for wf in grp.waveforms.values():
```
maarten-ic, line 197: Idem, no need to loop here

waveform_editor/configuration.py
```diff
@@ -139,8 +188,36 @@ def remove_group(self, path):
...
+                        raise RuntimeError(
+                            f"Cannot remove group {group.name}. "
+                            f"{wf.name!r} depends on waveform in it."
+                        )
+
+        del parent_group.groups[path[-1]]
         self._recursive_remove_waveforms(group)
+        self._calculate_bounds()
+        for wf in list(self.waveform_map.keys()):
+            if wf not in self.waveform_map:
```
maarten-ic, line 210: Couldn't you reuse the to_remove list?

waveform_editor/configuration.py
```diff
@@ -139,8 +188,36 @@ def remove_group(self, path):
...
+                self.dependency_graph.remove_node(wf)
+
+    def _collect_waveforms_in_group(self, group):
+        waveforms = set()
```
maarten-ic, line 214: Let's use a list here
```suggestion
        waveforms = []
```

maarten-ic (commented):
Quite some comments 😅 let me know if you have any questions :)

waveform_editor/gui/editor.py
SBlokhuizen: Something like the following would still work correctly, 
```
- {type: "linear", from: 10, to: 50, duration: 20}
- {'type': constant, value: 50, duration: 20}
- {type: linear, from: 50, 'to': 30, "duration": 20}
```

The following would not work though:
```
'- {type: linear, from: 10, to: 50, duration: 20}'
```

maarten-ic (commented):
Looks good to start making tests and docs!

docs/source/derived.rst
```diff
@@ -0,0 +1,111 @@
+.. _derived-waveforms:
```
maarten-ic, line 1: Can you add the example for NBI power/energy derived waveforms instead of only abstract examples?

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            self.is_constant = False
+            return node
+
+
+class DerivedWaveform(BaseWaveform):
+    def __init__(self, yaml_str, name, config, dd_version=None):
+        super().__init__(yaml_str, name, dd_version)
+        self.config = config
+        self.dependent_waveforms = set()
```
maarten-ic, line 72: Let's rename this to `self.dependencies`? _dependent waveforms_ sounds like they are dependent on us, while it's the other way around.

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+
+        try:
+            tree = ast.parse(str(self.yaml), mode="eval")
+        except Exception as e:
+            self.annotations.add(0, f"Could not parse or evaluate the waveform: {e}")
+            self.expression = None
+            return
+
+        extractor = ExpressionExtractor()
+        modified_tree = extractor.visit(ast.fix_missing_locations(tree))
```
maarten-ic, line 92: This should be executed the other way around, right? https://docs.python.org/3/library/ast.html#ast.NodeTransformer

```suggestion
        modified_tree = ast.fix_missing_locations(extractor.visit(tree))
```

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+
+    def __init__(self):
+        self.string_nodes = []
+        self.is_constant = True
+
+    def visit_Constant(self, node):
+        if isinstance(node.value, str):
+            self.string_nodes.append(node.value)
+            self.is_constant = False
```
maarten-ic, line 52: If `is_constant := not string_nodes` then we don't need the extra attribute

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+            Tuple containing the time and the derived waveform values.
+        """
+        if time is None:
+            # TODO: properly handle time for plotting
+            time = np.linspace(self.config.start, self.config.end, 1000)
+        if self.expression is None:
+            return time, np.zeros_like(time)
+
+        eval_context = self._build_eval_context(time)
+        aeval = Interpreter(user_symbols={"__w": eval_context}, minimal=True)
```
maarten-ic, line 147: I'm not sure yet if we should restrict access to non-vectorized numpy functions.

While `'test/1' / max('test/1')` works when giving the full time array, this will always return `1` when evaluating at specific times.
We could restrict the available context to only include numpy functions that operate element-wise. @DaanVanVugt how do you feel about this?

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+        with np.printoptions(threshold=10):
```
maarten-ic, line 148: Can you add a comment to explain why we do this?

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+            result = aeval.eval(self.expression, raise_errors=True)
+
+        # If derived waveform is a constant, ensure an array is returned
+        if self.is_constant:
+            return time, np.full_like(time, result)
+
+        # Ensure the result is a 1D array
+        if not isinstance(result, (list, np.ndarray)):
+            raise ValueError("Evaluation result is not a list or 1D array")
```
maarten-ic, line 157: Why would it be a list?

waveform_editor/derived_waveform.py
```diff
@@ -0,0 +1,170 @@
...
+        result = np.asarray(result)
+        if result.ndim != 1:
+            raise ValueError("Evaluation result is not 1-dimensional")
+        if result.shape[0] != len(time):
```
maarten-ic, line 161: I'd be happy to check `if result.shape != time.shape` with a single error message (that includes both shapes)

maarten-ic (commented):
Getting there! 🙂 

Mostly minor comments

waveform_editor/derived_waveform.py
SBlokhuizen: As discussed, let's only allow `np.func`s for now to mitigate this issue. This was fixed in the last commit.

waveform_editor/derived_waveform.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, yaml_str, name, config, dd_version=None):
         self.dependencies = set()
         self.is_constant = False
         self.expression = None
+        self.user_symbols = {
+            name: obj for name, obj in vars(np).items() if isinstance(obj, np.ufunc)
+        }
```
maarten-ic, line 72: Could we make this a constant class variable?

Also, let's iterate over public numpy API only using `__all__`?
```python
NUMPY_UFUNCS = {}
for name in np.__all__:
    obj = getattr(np, name)
    if isinstance(obj, np.ufunc):
        NUMPY_UFUNCS[name] = obj
```

maarten-ic (approved):
Nice work! Though in hindsight we should have split this PR in smaller bits :)

PR #78, opened 2025-07-03
add global exception handling

SBlokhuizen:
Adds a exception handler which:

- Logs the errors
- Creates a panel notification


maarten-ic:
LGTM, but can we merge this to develop instead of another PR?

PR #80, opened 2025-07-14
Fix textinput requiring two click in export dialog

SBlokhuizen:
This fixes the issue where textinputs in modals require two clicks to gain focus, using a solution that was proposed [here](https://discourse.holoviz.org/t/how-to-use-pn-modal-in-combination-with-a-template/8829). This fixes the issue for the `ExportModal`, but it does not solve it for modals which are triggered from a `ButtonIcon` object. This seems to be a bug in Panel, and is reported [here](https://github.com/holoviz/panel/issues/8015).

### DIFF
--- a/waveform_editor/gui/main.py
+++ b/waveform_editor/gui/main.py
@@ -98,7 +98,7 @@ class WaveformEditorGui(param.Parameterized):
         self.template = pn.template.FastListTemplate(
             title=f"Waveform Editor (v{waveform_editor.__version__})",
             main=self.tabs,
-            sidebar=sidebar,
+            sidebar=[sidebar],
             sidebar_width=400,
         )
 


waveform_editor/gui/main.py
```diff
@@ -98,7 +98,7 @@ def __init__(self):
         self.template = pn.template.FastListTemplate(
             title=f"Waveform Editor (v{waveform_editor.__version__})",
             main=self.tabs,
```
maarten-ic, line 100: Should main be a list as well? See the example listed in the docs: https://panel.holoviz.org/reference/templates/FastListTemplate.html#tips-tricks

maarten-ic (commented):
Interesting behaviour! But nice that this already fixes one use case :)

maarten-ic (approved):

PR #82, opened 2025-07-15
Show warning when discarding unsaved changes to YAML file

SBlokhuizen:
This implements a YAML save state. If the YAML file is changed, the user will be notified. If a new YAML file is created, an existing YAML file is loaded or exported, the user will be warned that there are unsaved changes.

https://github.com/user-attachments/assets/b37b3199-081b-4a34-963b-c1bf55087f6e


waveform_editor/gui/io/manager.py
```diff
@@ -55,27 +58,60 @@ def create_new_file(self):
...
     def _set_open_file_text(self):
+        alert = "⚠️ Unsaved changes " if self.main_gui.config.has_changed else ""
         if not self.is_editing:
             self.open_file_text.value = "Create or open a YAML file"
         elif self.open_file:
-            self.open_file_text.value = f"{self.open_file}"
+            self.open_file_text.value = f"{self.open_file}\n{alert}"
         else:
-            self.open_file_text.value = "Untitled_1.yaml"
+            self.open_file_text.value = f"Untitled_1.yaml\n{alert}"
```
maarten-ic, line 114: Can we improve this a bit?
image

1. Can we show the filename only, and maybe a tooltip for displaying the full path?
2. The "unsaved changes" is very present now. Maybe an icon (filled circle?) when there are unsaved changes, and another icon (checkmark?) or no icon at all when there are no unsaved changes? Again with a tooltip to explain the icon.

N.B. I'd be happy with a text field + one tooltip to give more details to cover both.

waveform_editor/gui/io/manager.py
```diff
@@ -55,27 +58,60 @@ def create_new_file(self):
...
     def _handle_menu_selection(self, event):
         clicked = event.new
         if clicked == NEW:
-            self.create_new_file()
+            self._confirm_and_execute(
+                self.create_new_file,
+                (
+                    "### ⚠️ **You have unsaved changes**  \n"
+                    "Opening a new file will discard all unsaved changes.  \n"
+                    "Do you want to continue?"
```
maarten-ic, line 78: DRY: this message is repeated three times with some differences. Can we do the following?

```python
# inside _confirm_unsaved_changes
message= f"""\
### ⚠️ **You have unsaved changes**
{message}  
Do you want to continue?"""


# Over here:
self._confirm_unsaved_changes(
    self.create_new_file,
    "Creating a new file will discard all unsaved changes"
)
```

waveform_editor/gui/io/manager.py
```diff
@@ -55,27 +58,60 @@ def create_new_file(self):
...
         self.open_file = None
 
+    def _confirm_and_execute(self, action, message):
+        def proceed():
+            action()
+
+        if self.main_gui.config.has_changed:
+            self.main_gui.confirm_modal.show(message, on_confirm=proceed)
+        else:
+            proceed()
```
maarten-ic, line 68: The following should work, no?

```suggestion
        if self.main_gui.config.has_changed:
            self.main_gui.confirm_modal.show(message, on_confirm=action)
        else:
            action()
```

waveform_editor/gui/io/manager.py
```diff
@@ -55,27 +58,60 @@ def create_new_file(self):
         self.main_gui.load_yaml(yaml_content)
         self.open_file = None
 
+    def _confirm_and_execute(self, action, message):
```
maarten-ic, line 61: Let's rename to `_confirm_unsaved_changes`, because it doesn't ask the user when there are no unsaved changes?

maarten-ic (commented):
Some suggestions for improvements:

PR #86, opened 2025-07-21
Add basic XML exporter for PCSSP

SBlokhuizen:
This adds support for exporting to a PCSSP XML file. 

Currently only very basic XML export is supported:
- Only a single segment
- All entry rules are set to `None`
- All exit rules are set to `Last`
- All execution rules are set to `Linear`
- Only signals with signal_type `Amplitude`, type `double` and dimension of 1 are supported

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -139,7 +139,7 @@ This command reads the waveform definitions from the `YAML` file, evaluates them
 .. note::
     You must provide exactly one of `--linspace` or `--csv` for this command.
 
-export-xml
+export-pcssp-xml
 ----------
 
 Exports the waveform data to a PCSSP-compatible XML file.
@@ -148,7 +148,7 @@ Exports the waveform data to a PCSSP-compatible XML file.
 
 .. code-block:: bash
 
-   waveform-editor export-xml [OPTIONS] YAML OUTPUT_XML
+   waveform-editor export-pcssp-xml [OPTIONS] YAML OUTPUT_XML
 
 **Description:**
 
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -139,5 +139,32 @@ This command reads the waveform definitions from the `YAML` file, evaluates them
 .. note::
     You must provide exactly one of `--linspace` or `--csv` for this command.
 
+export-xml
+----------
+
+Exports the waveform data to a PCSSP-compatible XML file.
+
+**Usage:**
+
+.. code-block:: bash
+
+   waveform-editor export-xml [OPTIONS] YAML OUTPUT_XML
+
+**Description:**
+
+This command reads waveform definitions from the given `YAML` file, evaluates them at the specified time points (via ``--linspace`` or ``--csv``), and exports the result to a PCSSP-compatible XML file at the path specified by ``OUTPUT_XML``. The XML format includes signal declarations and associated time-based trajectories.
+
+**Arguments:**
+
+*   ``YAML``: Path to the input waveform YAML configuration file.
+*   ``OUTPUT_XML``: Path to the file where the XML data will be saved. The parent directory will be created if it does not exist.
+
+**Options:**
+
+*   ``--linspace START,STOP,NUM``: Define time points using `numpy.linspace`. (See `Specifying Time Points for Export`_).
+*   ``--csv PATH``: Define time points using a CSV file. (See `Specifying Time Points for Export`_).
+
+.. note::
+    You must provide exactly one of `--linspace` or `--csv` for this command.
 
 .. _IMAS: https://imas.iter.org/
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -140,7 +140,7 @@ This command reads the waveform definitions from the `YAML` file, evaluates them
     You must provide exactly one of `--linspace` or `--csv` for this command.
 
 export-pcssp-xml
-----------
+----------------
 
 Exports the waveform data to a PCSSP-compatible XML file.
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -100,11 +100,11 @@ def test_export_csv(runner, tmp_path, test_yaml_file, test_csv_file):
 
 def test_export_xml(runner, tmp_path, test_yaml_file, test_csv_file):
     csv_path, _ = test_csv_file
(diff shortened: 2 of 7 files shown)


docs/source/cli.rst
```diff
@@ -139,5 +139,32 @@ This command reads the waveform definitions from the `YAML` file, evaluates them
 .. note::
     You must provide exactly one of `--linspace` or `--csv` for this command.
 
+export-xml
```
maarten-ic, line 142: Please name the PCSSP export as such (XML export is too generic) :)

Also holds for the code, classes, etc.

tests/test_cli.py
```diff
@@ -98,6 +98,23 @@ def test_export_csv(runner, tmp_path, test_yaml_file, test_csv_file):
     assert output_csv.exists()
 
 
+def test_export_xml(runner, tmp_path, test_yaml_file, test_csv_file):
+    csv_path, _ = test_csv_file
+    output_xml = tmp_path / "test.csv"
```
maarten-ic, line 103:
```suggestion
    output_xml = tmp_path / "test.xml"
```

waveform_editor/exporter.py
```diff
@@ -24,6 +26,18 @@ def __init__(self, config, times, progress=None):
         if self.times is not None and not np.all(np.diff(self.times) > 0):
             raise ValueError("Time array must be in increasing order.")
 
+    def to_xml(self, file_path):
```
maarten-ic, line 29:
```suggestion
    def to_pcssp_xml(self, file_path):
```

waveform_editor/exporter.py
```diff
@@ -24,6 +26,18 @@ def __init__(self, config, times, progress=None):
...
+        """Export the configuration to a PCSSP XML file.
+
+        Args:
+            file_path: The file path to store the XML file to.
+        """
+        xml_exporter = XMLExporter(self.config, self.times)
```
maarten-ic, line 35:
```suggestion
        xml_exporter = PCSSPExporter(self.config, self.times)
```

waveform_editor/gui/io/file_exporter.py
```diff
@@ -17,6 +17,7 @@
 IDS_EXPORT = "IDS"
 CSV_EXPORT = "CSV"
 PNG_EXPORT = "PNG"
+XML_EXPORT = "XML"
```
maarten-ic, line 20:
```suggestion
PCSSP_EXPORT = "XML"
```

waveform_editor/xml_exporter.py
```diff
@@ -0,0 +1,75 @@
...
+        outputs = ET.SubElement(declarations, "OUTPUTS")
+
+        self._add_signals(outputs)
+
+        segments = ET.SubElement(root, "SEGMENTS")
+        # Only a single segment is currently supported
+        segment = ET.SubElement(
+            segments,
+            "SEGMENT",
+            # Watchdogs are unused, so left empty
```
maarten-ic, line 28: @timoravensbergen Do we need to fill something in for the `wd_time` and `wd_target` in the `<SEGMENT>`  elements? Can we leave them empty (as current), or leave them away altogether?

timoravensbergen: For the watchdog time-out I suggest to put the total length of the segment. In most cases for us reaching the end of a segment means something is not right and we switch to soft event handling. In case the exit condition is simply time-based, we simply would switch to the next segment. For the watchdog target I would put `EHTerm1` which is our default soft EH.

SBlokhuizen:
Hi @timoravensbergen, below I have created a small example containing two waveforms that was exported to a PCSSP XML. Could you validate that this is the correct format for the PCSSP?
<details>

<summary>YAML waveform configuration used to generate image</summary>

```YAML
currents:
  core_profiles/global_quantities/ip:
  - {from: 0, to: 1e5}
  - {duration: 2}
  - {to: 0}
power:
  ec_launchers/beam(1)/power_launched:
  - {from: 0, to: 5e5}
  - {duration: 2}
  - {to: 0}
```

</details>
 
image

Output of the PCSSP XML exporter:
```XML
<?xml version='1.0' encoding='utf-8'?>
<SCHEDULE>
  <DECLARATIONS>
    <PARAMETERS />
    <OUTPUTS>
      <SIGNAL name="core_profiles/global_quantities/ip" signal_type="Amplitude" type="double" dimension="1" description="Total plasma current (toroidal component). Positive sign means anti-clockwise when viewed from above." value="0" />
      <SIGNAL name="ec_launchers/beam(1)/power_launched" signal_type="Amplitude" type="double" dimension="1" description="Beam power launched into the vacuum vessel" value="0" />
    </OUTPUTS>
  </DECLARATIONS>
  <SEGMENTS>
    <SEGMENT name="Waveform Editor" id="0" wd_time="" wd_target="">
      <SIGNALS_TRAJECTORIES>
        <SIGNAL_TRAJECTORY name="core_profiles/global_quantities/ip">
          <ENTRY_RULE is="None" />
          <EXECUTION_RULE is="Linear" />
          <EXIT_RULE is="Last" />
          <REFERENCE>
            <POINT time="0.0" value="0.0" />
            <POINT time="1.0" value="100000.0" />
            <POINT time="3.0" value="100000.0" />
            <POINT time="4.0" value="0.0" />
          </REFERENCE>
        </SIGNAL_TRAJECTORY>
        <SIGNAL_TRAJECTORY name="ec_launchers/beam(1)/power_launched">
          <ENTRY_RULE is="None" />
          <EXECUTION_RULE is="Linear" />
          <EXIT_RULE is="Last" />
          <REFERENCE>
            <POINT time="0.0" value="0.0" />
            <POINT time="1.0" value="500000.0" />
            <POINT time="3.0" value="500000.0" />
            <POINT time="4.0" value="0.0" />
          </REFERENCE>
        </SIGNAL_TRAJECTORY>
      </SIGNALS_TRAJECTORIES>
    </SEGMENT>
  </SEGMENTS>
</SCHEDULE>

```

waveform_editor/xml_exporter.py
timoravensbergen: Hmm in our case the segment always starts at t=0, it is relative to the start of that particular segment (and not absolute pulse time). In that case I would maybe put your `end time` which is when our watchdog would need triggering to say something is off

maarten-ic (approved):
LGTM 🚀

PR #87, opened 2025-07-22
Dynamic Options dialog

SBlokhuizen:
Adds options parameterized classes which support loading/saving to a YAML file, and can be used for dynamic Panel UI creation.


waveform_editor/options.py
```diff
@@ -0,0 +1,93 @@
...
+from pathlib import Path
+
+import panel as pn
+import param
+import yaml
+
+logger = logging.getLogger(__name__)
+
+_xdg = os.environ.get("XDG_CONFIG_HOME")
+_config_home = Path(_xdg) if _xdg else Path(os.environ["HOME"]) / ".config"
```
maarten-ic, line 12: `HOME` may be unset, let's use https://docs.python.org/3/library/pathlib.html#pathlib.Path.home
image

waveform_editor/options.py
```diff
@@ -0,0 +1,93 @@
...
+    def _save_settings(self, event=None):
+        """Serialize current configuration to disk in YAML format."""
+        config = {
+            p: getattr(self, p) for p in self.param if p != "name" and p != "nice"
+        }
+
+        if self.gs_solver == "NICE":
+            config["nice"] = self.nice.to_dict()
+
+        with open(CONFIG_FILE, "w") as f:
```
maarten-ic, line 81: This will fail if the `_config_home` directory doesn't exist:

```python
>>> import pathlib
>>> open(pathlib.Path("/tmp/.config/test.yaml"), "w")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/.config/test.yaml'
```

waveform_editor/options.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class NiceOptions(param.Parameterized):
+    executable = param.String(
+        label="NICE executable path", doc="Path to NICE inverse IMAS MUSCLE3 executable"
+    )
+    environment = param.Dict(
+        label="NICE environment variables", doc="Environment variables for NICE"
+    )
+    modules = param.List(label="NICE modules", doc="EasyBuild Modules to load for NICE")
```
maarten-ic, line 23:
```suggestion
    modules = param.List(label="NICE modules", doc="Environment Modules to load for NICE")
```

maarten-ic (commented):
Nice! See comments below

maarten-ic (approved):
Thanks!

PR #88, opened 2025-07-23
Fix constant constant waveforms not showing in plot

### LINKED ISSUE #84: Constant waveforms disappear from plot
https://github.com/user-attachments/assets/98314714-51cb-4188-b31c-e9cb48c66a52

SBlokhuizen:
Fixes #84. Derived waveforms run from the `start` to the `end` of the configuration, which are both set to 0 by default. As constant values are now also derived waveforms, they also run from the `start` to the `end` of the configuration, which means they run from 0 to 0 if there are no other waveforms, which caused them to not display. To fix this, the default end of the configuration was set to 1 instead.

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -360,4 +360,5 @@ def test_bounds(config):
     assert config.end == 10
 
     config.remove_group(path)
-    assert config.start == config.end == 0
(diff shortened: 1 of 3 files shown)


maarten-ic (approved):
👍

PR #89, opened 2025-07-24
Feature/yaml globals

### LINKED ISSUE #71: Unable to add/change globals dictionary from GUI
You cannot yet set the globals dictionary from the GUI. It would be nice if you can change the machine description / DD version from there.

SBlokhuizen:
Fixes #71. This introduces a `YamlGlobals` class responsible for managing the global YAML parameters. The global parameters can be changed through the `YamlGlobalsDialog` which dynamically builds a UI of the `YamlGlobals` class.

https://github.com/user-attachments/assets/02234876-161d-4d06-9a13-cda10674a8d3

### DIFF
--- a/waveform_editor/gui/globals_dialog.py
+++ b/waveform_editor/gui/globals_dialog.py
@@ -1,0 +1,31 @@
+import panel as pn
+import param
+from panel.viewable import Viewer
+
+
+class YamlGlobalsDialog(Viewer):
+    visible = param.Boolean(allow_refs=True)
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+        self.modal = pn.Modal(pn.Param(self.config.globals, name="YAML Globals"))
+
+        self.button = pn.widgets.ButtonIcon(
+            icon="settings",
+            size="20px",
+            active_icon="settings-filled",
+            description="Change YAML global parameters",
+            on_click=lambda event: self.modal.show(),
+            visible=self.param.visible,
+        )
+
+        for param_name in self.config.globals.param:
+            if param_name != "name":
+                self.config.globals.param.watch(self._on_param_change, param_name)
+
+    def _on_param_change(self, event):
+        self.config.has_changed = True
+
+    def __panel__(self):
+        return pn.Column(self.button, self.modal)
--- a/waveform_editor/gui/globals_dialog.py
+++ b/waveform_editor/gui/globals_dialog.py
@@ -20,6 +20,7 @@ class YamlGlobalsDialog(Viewer):
             visible=self.param.visible,
         )
 
+        # Trigger configuration's has_changed boolean when a global param is changed
         for param_name in self.config.globals.param:
             if param_name != "name":
                 self.config.globals.param.watch(self._on_param_change, param_name)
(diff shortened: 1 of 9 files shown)


waveform_editor/gui/globals_dialog.py
```diff
@@ -0,0 +1,32 @@
...
+from panel.viewable import Viewer
+
+
+class YamlGlobalsDialog(Viewer):
+    visible = param.Boolean(allow_refs=True)
+
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+        self.modal = pn.Modal(pn.Param(self.config.globals, name="YAML Globals"))
```
maarten-ic, line 12: The GUI element for MD URIs is not very clear:
image

Can we use a different widget? Looking at the panel gallery perhaps the [Tabulator](https://panel.holoviz.org/reference/widgets/Tabulator.html) is the easiest way to get a nice UI?

waveform_editor/gui/globals_dialog.py
```diff
@@ -0,0 +1,32 @@
...
+
+        self.button = pn.widgets.ButtonIcon(
```
maarten-ic, line 14: This button looks free floating in the application and it's not clear that this will edit the global properties of the current waveform configuration:
image

- The gear icon suggests it's for changing settings, but that's not really what this one is for
- Maybe we can put it on the same row as the File menu and filename?

waveform_editor/gui/main.py
```diff
@@ -50,12 +51,15 @@ def __init__(self):
         self.confirm_modal = ConfirmModal()
         self.rename_modal = RenameModal()
         self.io_manager = IOManager(self)
+        globals_dialog = YamlGlobalsDialog(self.config)
+        globals_dialog.visible = self.io_manager.param.is_editing.rx.bool()
```
maarten-ic, line 55: Why don't we just always start with an empty configuration (like you start with an empty file with notepad) and get rid of this `is_editing` flag?

SBlokhuizen: Good idea! I have changed this in the last commit.

waveform_editor/dict_editor.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __init__(self, key_options=None, names=("key", "value"), **params):
+        """Initialize the DictEditor widget.
+
+        Args:
+            key_options: List of allowed keys shown in the dropdown editor.
+            names: Tuple of column names to display for keys and values.
+        """
+        if key_options is None:
+            key_options = []
```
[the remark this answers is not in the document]
SBlokhuizen: We shouldn't use mutable objects are arguments, as they are only evaluated once. Ruff actually prevents you from doing this as well:
https://docs.astral.sh/ruff/rules/mutable-argument-default/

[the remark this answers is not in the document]
maarten-ic: That's true, but we're using it as a readonly variable, so this doesn't really apply.

Maybe we should just not put a default arg at all, since our only use case provides an argument anyway?

maarten-ic (approved):
Please test my latest commit :)

SBlokhuizen:
> Please test my latest commit :)

Seems to work well on my end!

PR #93, opened 2025-08-01
initial UI for NICE equilibrium reconstruction

SBlokhuizen:
Implements plasma shape editor tab in the main UI, with which you can send equilibrium reconstruction jobs to NICE and visualize the results. 

[Screencast from 2025-08-01 09-52-14.webm](https://github.com/user-attachments/assets/cbc37aea-9872-4787-9f7c-f65f25d4ed4a)


SBlokhuizen:
@maarten-ic I have updated it so the IDSs are automatically loaded once  NICE settings are changed. Could you try it again? Thanks

waveform_editor/shape_editor/nice_plotter.py
```diff
@@ -0,0 +1,219 @@
...
+    HEIGHT = 1000
+
+    def __init__(self, communicator, plotting_params, **params):
+        super().__init__(**params)
+        self.communicator = communicator
+        self.plotting_params = plotting_params
+        self.wall = None
+        self.pf_active = None
+
+    @pn.depends("plotting_params.param", "communicator.processing")
```
maarten-ic, line 30: Would it be possible to construct several DynamicMaps for the different components?

Right now there is a ~second delay to just turn on/off display of e.g. coils (probably because it's recalculating all of the contours).

waveform_editor/shape_editor/nice_plotter.py
```diff
@@ -0,0 +1,219 @@
...
+                    z0 = rect.z - rect.height / 2
+                    z1 = rect.z + rect.height / 2
+                    rectangles.append((r0, z0, r1, z1))
+                elif outline.has_value:
+                    path = hv.Path([list(zip(outline.r, outline.z))]).opts(
+                        color="black", line_width=1, show_legend=False
+                    )
+                    paths.append(path)
+                else:
+                    continue
```
maarten-ic, line 95: Can you add a log statement that you skipped plotting an element?

waveform_editor/shape_editor/nice_plotter.py
```diff
@@ -0,0 +1,219 @@
...
+                    r0 = rect.r - rect.width / 2
+                    r1 = rect.r + rect.width / 2
+                    z0 = rect.z - rect.height / 2
+                    z1 = rect.z + rect.height / 2
+                    rectangles.append((r0, z0, r1, z1))
+                elif outline.has_value:
+                    path = hv.Path([list(zip(outline.r, outline.z))]).opts(
+                        color="black", line_width=1, show_legend=False
+                    )
+                    paths.append(path)
```
maarten-ic, line 93: Can you make rectangles and path logic consistent?

E.g.
```
paths.append((outline.r, outline.z))
[...]
paths = hv.Path(paths).opts(
    color="black", line_width=1, show_legend=False
)
```

waveform_editor/shape_editor/nice_plotter.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Args:
+            equilibrium: an equilibrium IDS.
+
+        Returns:
+            Holoviews curve containing the separatrix.
+        """
+        r = equilibrium.time_slice[0].boundary.outline.r
+        z = equilibrium.time_slice[0].boundary.outline.z
+        separatrix = hv.Curve((r, z)).opts(color="red", line_width=2, show_legend=False)
```
maarten-ic, line 154: Any reason that you use a Curve here and a Path for the coil outlines?

waveform_editor/shape_editor/plotting_params.py
```diff
@@ -0,0 +1,17 @@
+import param
+
+
+class PlottingParams(param.Parameterized):
```
maarten-ic, line 4: Would it make more sense to put these as parameters on the NicePlotter object?

waveform_editor/shape_editor/nice_plotter.py
maarten-ic: As discussed, see #95

PR #96, opened 2025-08-04
Allow to interrupt `waveform-editor gui` with Ctrl+C

maarten-ic:
And construct only a single WaveformEditorGui instance.


SBlokhuizen (approved):
LGTM!

PR #99, opened 2025-08-05
Do not show save waveform button if no waveform is selected

SBlokhuizen:
Previously, clicking the save waveform button would cause an error if no waveform was being edited. Now, the save button only appears when a waveform is actively being edited.

### DIFF
--- a/waveform_editor/gui/editor.py
+++ b/waveform_editor/gui/editor.py
@@ -15,9 +15,10 @@ class WaveformEditor(Viewer):
         class_=(Waveform, DerivedWaveform),
         doc="Waveform currently being edited. Use `set_waveform` to change.",
     )
-    code_editor = param.ClassSelector(class_=pn.widgets.CodeEditor)
     error_alert = param.ClassSelector(class_=pn.pane.Alert, default=pn.pane.Alert())
-    has_changed = param.Boolean(doc="Whether there are unsaved changes in the editor.")
+    has_changed = param.Boolean(
+        allow_refs=True, doc="Whether there are unsaved changes in the editor."
+    )
     stored_string = param.String(
         default=None,
         doc="Contains the waveform text before any changes were made in the editor.",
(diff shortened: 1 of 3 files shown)


waveform_editor/gui/editor.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, config):
             active_icon="check",
             description="Save waveform",
             on_click=self.save_waveform,
+            visible=self.param.waveform.rx.pipe(bool),
```
maarten-ic, line 43: Would be even nicer if we could make it depend on self.has_changed (but that one is not declarative) :)

maarten-ic (approved):

maarten-ic (commented):
Nice improvements, I've got two comments, let me know if you don't agree with them.

waveform_editor/gui/editor.py
```diff
@@ -15,21 +15,24 @@ class WaveformEditor(Viewer):
         class_=(Waveform, DerivedWaveform),
         doc="Waveform currently being edited. Use `set_waveform` to change.",
     )
-    code_editor = param.ClassSelector(class_=pn.widgets.CodeEditor)
-    error_alert = param.ClassSelector(class_=pn.pane.Alert, default=pn.pane.Alert())
-    has_changed = param.Boolean(doc="Whether there are unsaved changes in the editor.")
```
maarten-ic: Why not keep this as a param with `allow_refs=True`?

That's nicer for downstream code, who would otherwise need to do `editor.has_changed.rx.value` to get the value.

waveform_editor/gui/editor.py
```diff
@@ -90,55 +93,39 @@ def on_value_change(self, event):
             waveform_yaml = f"{name}: {editor_text}"
         waveform = self.config.parse_waveform(waveform_yaml)
         self.handle_exceptions(waveform)
-        if not self.error_alert.object:
+        if not self.param.error_message.rx.bool().rx.value:
```
maarten-ic, line 96:
```suggestion
        if not self.error_message:
```

waveform_editor/gui/io/manager.py
```diff
@@ -57,7 +57,7 @@ def create_new_file(self):
         self.open_file = None
 
     def _confirm_unsaved_changes(self, action, message):
-        if self.main_gui.config.has_changed or self.main_gui.editor.has_changed():
+        if self.main_gui.config.has_changed or self.main_gui.editor.has_changed:
```
maarten-ic, line 60: `editor.has_changed` is a reactive expression, which will always evaluate True-ish

Let's make it a param again so this code keeps working :)

PR #112, opened 2025-08-11
Allow slashes in both waveform names and group names

SBlokhuizen:


maarten-ic (approved):

maarten-ic (changes requested):
Can you update the documentation as well? https://waveform-editor.readthedocs.io/en/latest/yaml_format.html#grouping-waveforms

maarten-ic (approved):

PR #115, opened 2025-08-13
Fix NICE coil currents from GUI 

SBlokhuizen:
Allows setting the currents of the coil to a fixed value from the UI. 

[Screencast from 2025-08-13 14-55-57.webm](https://github.com/user-attachments/assets/85682ce0-baa1-44b3-a41c-11d426ae7048)

Note: NICE allows for linking multiple coils such that they must always have the same current. This is not visualized in the UI. The ITER defaults of NICE are used here, where the CS1L and CS1U coils must have the same current. If you want to fix the current for these coils, you need to fix the current for CS1U (fixing the coil for CS1L will not have any effect).

### DIFF
--- a/waveform_editor/shape_editor/coil_currents.py
+++ b/waveform_editor/shape_editor/coil_currents.py
@@ -3,7 +3,6 @@ import xml.etree.ElementTree as ET
 import numpy as np
 import panel as pn
 import param
-from bokeh.models.formatters import PrintfTickFormatter
 from panel.viewable import Viewer
 
 
@@ -52,7 +51,7 @@ class CoilCurrents(Viewer):
                 start=-5e4,
                 end=5e4,
                 disabled=checkbox.param.value.rx.not_(),
-                format=PrintfTickFormatter(format=f"%.0f"),
+                format="0",
                 width=450,
             )
             row = pn.Row(checkbox, slider)


waveform_editor/shape_editor/coil_currents.py
```diff
@@ -0,0 +1,143 @@
...
+        Args:
+            xml_root: XML root element of the NICE parameters.
+            params: Dictionary mapping XML element names to new values.
+
+        Returns:
+            Updated XML as a string.
+        """
+        for key, val in params.items():
+            elem = xml_root.find(key)
+            if elem is not None:
```
maarten-ic, line 138: What if elem is None? Shouldn't that be an error?

waveform_editor/shape_editor/coil_currents.py
```diff
@@ -0,0 +1,143 @@
...
+            group_fixed_index = "-1"
+        else:
+            group_fixed_index = " ".join(map(str, fixed_groups))
+
+        params = {
+            "n_group_fixed_index": n_group_fixed_index,
+            "group_fixed_index": group_fixed_index,
+        }
+
+        return self._update_xml_params(root, params)
```
maarten-ic, line 124: This seems a bit overcomplicated, why not directly update the XML here and get rid of the `_update_xml_params` method?

```suggestion
        root.find("n_group_fixed_index").text = str(len(fixed_groups))
        # NICE requires group_fixed_index to be filled even when there are no fixed coils
        root.find("group_fixed_index").text = " ".join(fixed_groups) or "-1"

        return rootET.tostring(root, encoding="unicode")
```

waveform_editor/shape_editor/coil_currents.py
```diff
@@ -0,0 +1,143 @@
...
+            Updated XML string with the `n_group_fixed_index` and `group_fixed_index`
+                elements modified.
+        """
+        root = ET.fromstring(xml_string)
+        coil_groups = list(map(int, root.find("coil_group_index").text.split()))
+
+        fixed_coils = [i for i, row in enumerate(self.coil_ui) if row.objects[0].value]
+
+        target_groups = {coil_groups[coil_idx] for coil_idx in fixed_coils}
+        fixed_groups = sorted(list(target_groups))
```
maarten-ic, line 109: No need to map coil_groups to an int (and later map them back to str):

```suggestion
        coil_groups = root.find("coil_group_index").text.split()

        fixed_coils = [i for i, row in enumerate(self.coil_ui) if row.objects[0].value]

        target_groups = {coil_groups[coil_idx] for coil_idx in fixed_coils}
        fixed_groups = sorted(list(target_groups), key=int)
```

waveform_editor/shape_editor/coil_currents.py
```diff
@@ -0,0 +1,143 @@
...
+            coil_current = coil.current
+            checkbox = pn.widgets.Checkbox(value=False, margin=(30, 10, 10, 10))
+            slider = pn.widgets.EditableFloatSlider(
+                name=f"{coil.name}",
+                value=coil_current.data[0],
+                start=-5e4,
+                end=5e4,
+                disabled=True,
+                format=PrintfTickFormatter(
+                    format=f"%.0f {coil_current.metadata.units}"
```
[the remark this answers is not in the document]
maarten-ic: NaN's are infectious, and after running NICE once with a NaN it's impossible to get the shape editor back into a shape where it will work 🙈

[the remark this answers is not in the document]
SBlokhuizen: Regarding the first point, I have moved the units into the name of the slider, that should it. 

The second point seems to be an issue with panel, I have created a report [here](https://github.com/holoviz/panel/issues/8123).

maarten-ic (approved):

PR #117, opened 2025-08-14
Parameterisation of p' and ff' profiles

SBlokhuizen:
`PlasmaProperties` now supports manual input. The p' and ff' profiles can be parameterized using alpha, beta, and gamma parameters.

[Screencast from 2025-08-14 15-25-38.webm](https://github.com/user-attachments/assets/27923b8a-72a6-4f9d-b6ea-0d1e3bfd8ff8)


maarten-ic (commented):
I like the simplifications :) Couple of comments, some already discussed

maarten-ic (approved):

PR #122, opened 2025-08-19
Create docs explaining the GUI

SBlokhuizen:
Adds docs page for the GUI. The part for the plasma shape editor is still left as TODO, as this is still being worked on.

### DIFF
--- a/docs/source/gui.rst
+++ b/docs/source/gui.rst
@@ -22,22 +22,25 @@ The GUI can be launched using the following command in your terminal:
 
     waveform-editor gui
 
-This normally opens automatically in your default web browser. If it does not, 
-you can manually open the app in a web browser by going to the address printed in the terminal.
-For example, when the ``waveform-editor gui`` command has the following output, the app is running
-on the web address ``http://localhost:38895``.
-
-.. code-block:: console
-
-    $ waveform-editor gui
-    [...]
-    Launching server at http://localhost:38895
 
 You can also optionally specify a YAML file to load on startup:
 
 .. code-block:: bash
 
     waveform-editor gui /path/to/your/waveforms.yaml
+
+.. note::
+
+  This normally opens automatically in your default web browser. If it does not, 
+  you can manually open the app in a web browser by going to the address printed in the terminal.
+  For example, when the ``waveform-editor gui`` command has the following output, the app is running
+  on the web address ``http://localhost:38895``.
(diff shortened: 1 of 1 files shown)


docs/source/gui.rst
```diff
@@ -6,16 +6,213 @@ Graphical User Interface
...
+* Edit waveforms through a text editor with live plot updates and error
+  highlighting.
+* Export waveform data to various formats, including IDS, CSV, PNG, and PCSSP XML.
+* Graphically edit and design the desired plasma shape using NICE.
 
-The GUI can be launched using:
+Launching the GUI
+-----------------
+
+The GUI can be launched using the following command in your terminal:
```
maarten-ic, line 19: Depending on the user's configuration, it could be that the webbrowser doesn't launch automatically. Can you add something here like:

- This should open a webbrowser
- If not, the command will print the web address (localhost:12345) where the app is running. You can open a webbrowser and go there yourself

docs/source/gui.rst
```diff
@@ -22,6 +22,9 @@ The GUI can be launched using the following command in your terminal:
 
     waveform-editor gui
 
+This normally opens automatically in your default web browser. If it does not, 
+you can manually enter the address printed in the terminal (e.g. ``http://localhost:12345``) to access the app.
```
maarten-ic, line 26:
```suggestion
you can manually open the app in a web browser by going to the address printed in the terminal.
For example, when the ``waveform-editor gui`` command has the following output, the app is running
on the web address ``http://localhost:38895``.

.. code-block:: console

    $ waveform-editor gui
    [...]
    Launching server at http://localhost:38895
```

docs/source/gui.rst
```diff
@@ -6,16 +6,224 @@ Graphical User Interface
...
+This normally opens automatically in your default web browser. If it does not, 
+you can manually open the app in a web browser by going to the address printed in the terminal.
+For example, when the ``waveform-editor gui`` command has the following output, the app is running
+on the web address ``http://localhost:38895``.
+
+.. code-block:: console
+
+    $ waveform-editor gui
+    [...]
+    Launching server at http://localhost:38895
```
maarten-ic, line 34: Let's put this in a `.. note::`? That probably looks nicer in the generated docs :)

maarten-ic (approved):

PR #124, opened 2025-08-28
Implement NICE direct mode

SBlokhuizen:
This introduces support for running NICE in direct mode from the plasma shape editor. A new radio button allows switching between NICE inverse and direct modes. To use direct mode, the path to the NICE MUSCLE3 executable must be provided.

Coil currents from the most recent run are preserved, enabling a workflow where NICE is first run in inverse mode to estimate coil currents, followed by manual adjustments in direct mode for higher precision.

Note: NICE direct mode exhibits nondeterministic behavior. Repeated runs with identical settings do not always produce consistent results—starting from coil currents obtained via inverse mode may sometimes converge to the expected shape, but in other cases may fail to converge.


waveform_editor/shape_editor/coil_currents.py
```diff
@@ -49,19 +49,13 @@ def create_ui(self, pf_active):
             coil_current = coil.current
             checkbox = pn.widgets.Checkbox(
                 margin=(30, 10, 10, 10),
-                disabled=self.nice_settings.param.mode.rx()
-                == self.nice_settings.DIRECT_MODE,
+                disabled=self.nice_settings.param.is_direct_mode.rx(),
```
maarten-ic, line 52: This can directly use the param, right?

```suggestion
                disabled=self.nice_settings.param.is_direct_mode,
```

waveform_editor/settings.py
```diff
@@ -47,6 +47,17 @@ class NiceSettings(param.Parameterized):
...
+    is_inverse_mode = param.Boolean(precedence=-1)
+
+    @param.depends("mode", watch=True, on_init=True)
+    def set_mode_flags(self):
+        if self.mode == self.INVERSE_MODE:
+            self.is_direct_mode = False
+            self.is_inverse_mode = True
+        else:
+            self.is_direct_mode = True
+            self.is_inverse_mode = False
```
maarten-ic, line 60: While the previous logic is correct, this better conveys intention:

```suggestion
        self.is_direct_mode = self.mode == self.DIRECT_MODE
        self.is_inverse_mode = self.mode == self.INVERSE_MODE
```

maarten-ic (approved):

PR #127, opened 2025-09-26
Option to store coil currents values in waveforms 

SBlokhuizen:
Adds a button in the coil current option to store the current coil current values in waveforms, as piecewise linear tendencies.

### DIFF
--- a/waveform_editor/tendencies/base.py
+++ b/waveform_editor/tendencies/base.py
@@ -50,7 +50,6 @@ class BaseTendency(param.Parameterized):
     duration = param.Number(
         default=1.0,
         bounds=(0.0, None),
-        inclusive_bounds=(False, True),
         doc="The duration of the tendency.",
     )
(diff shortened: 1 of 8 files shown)


waveform_editor/tendencies/base.py
```diff
@@ -79,9 +78,10 @@ class BaseTendency(param.Parameterized):
     )
     annotations = param.ClassSelector(class_=Annotations, default=Annotations())
 
-    def __init__(self, **kwargs):
+    def __init__(self, allow_zero_duration=False, **kwargs):
         self.line_number = kwargs.pop("line_number", 0)
         self.is_first_repeated = kwargs.pop("is_first_repeated", False)
+        self.allow_zero_duration = allow_zero_duration
```
maarten-ic, line 84: `allow_zero_duration` is a class property, so we can just make this a class variable instead of an instance variable?

maarten-ic (approved):

PR #132, opened 2025-10-29
Fix dark mode switch

SBlokhuizen:
Applied proposed solution addressed in [this issue](https://github.com/holoviz/panel/issues/8251).


ioan-alexandra (approved):

maarten-ic (approved):

maarten-ic:
Can you please check if the configuration is discarded when switching between dark/light mode?

ioan-alexandra:
for me dark mode already works now on the application only issue i see is that the graph lines are not dark-mode friendly but thats not related to this pr from what i see. close?

image

PR #133, opened 2025-10-29
Fix ruff linting

SBlokhuizen:
Ruff seems to want some new code style changes.

I just let ruff autofix them using:
```bash
ruff check waveform_editor --fix
```

### DIFF
--- a/waveform_editor/exporter.py
+++ b/waveform_editor/exporter.py
@@ -193,7 +193,9 @@ class ConfigurationExporter:
         # Here, phase/angle should be filled for all 4 beams.
         # However, certain niche cases involving multiple slices for different waveforms
         # might still not be handled correctly.
-        for waveform, (path, values) in zip(waveforms, values_per_waveform):
+        for waveform, (path, values) in zip(
+            waveforms, values_per_waveform, strict=False
+        ):
             logger.debug(f"Filling {waveform.name}...")
             self._fill_nodes_recursively(ids, path, values)
             self._increment_progress()
@@ -228,7 +230,7 @@ class ConfigurationExporter:
             if node.metadata.type.is_dynamic and part != path.parts[-1]:
                 if len(node) != len(values):
                     node.resize(len(values), keep=True)
-                for item, value in zip(node, values):
+                for item, value in zip(node, values, strict=False):
                     self._fill_nodes_recursively(item, path, value, next_index)
             else:
                 self._fill_nodes_recursively(node, path, values, next_index)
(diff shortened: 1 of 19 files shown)


waveform_editor/gui/selector/selection_group.py
```diff
@@ -20,7 +20,7 @@ class SelectionGroup(Viewer):
     def __init__(
         self,
         selector: "WaveformSelector",
-        group: Union[WaveformConfiguration, WaveformGroup],
+        group: WaveformConfiguration | WaveformGroup,
```
maarten-ic, line 23: Though technically Python 3.9 will be EOL in two days, I'd prefer keeping the 3.9-compatible syntax here (HPC environments can be slow in updating python versions).

Otherwise we should be clear and specify that we require at least Python 3.10 (in the pyproject.toml)

SBlokhuizen: It seems that the panel version that we require (`v1.6.3`), [requires at least Python 3.10](https://pypi.org/project/panel/1.6.3/) either way. I have updated our pyproject.toml to reflect this.

waveform_editor/gui/io/filebrowser.py
```diff
@@ -147,7 +147,7 @@ def _update_files(
             paths.insert(0, "..")
             abbreviated.insert(0, "⬆ ..")
 
-        options = dict(zip(abbreviated, paths))
+        options = dict(zip(abbreviated, paths, strict=False))
```
maarten-ic:
```suggestion
        options = dict(zip(abbreviated, paths, strict=True))
```

`abbreviated` and `paths` are guaranteed to have the same length (with above logic) `strict=True` captures this

waveform_editor/exporter.py
```diff
@@ -193,7 +193,9 @@ def _fill_waveforms(self, ids, waveforms):
         # Here, phase/angle should be filled for all 4 beams.
         # However, certain niche cases involving multiple slices for different waveforms
         # might still not be handled correctly.
-        for waveform, (path, values) in zip(waveforms, values_per_waveform):
+        for waveform, (path, values) in zip(
+            waveforms, values_per_waveform, strict=False
```
maarten-ic: Idem, this should be `strict=True`, shouldn't it?

maarten-ic: Please review the other changes of `zip` as well. I'd expect we should use strict=True in all of them 🙂

maarten-ic (approved):

PR #138, opened 2026-03-30
reload NICE XML config every time NICE is run

SBlokhuizen:


SBlokhuizen:
To be implented in the future based on the plasma shape editor redesign.

PR #140, opened 2026-04-14
Chore/restructure files

ioan-alexandra:
restructuring some files to have better separation between panel/viewers and logic

- `waveform_editor/export/` — moved exporter.py and pcssp_exporter.py here
- `waveform_editor/yaml/ `— moved yaml_globals.py and yaml_parser.py here
- `waveform_editor/gui/` — moved dict_editor.py here; moved GUI-specific shape editor panels (coil_currents, nice_plotter, plasma_properties, plasma_shape) into gui/shape_editor/
- `waveform_editor/shape_editor/` — now contains only non-GUI logic: plasma_shape_calc.py, plasma_properties_calc.py, nice_integration.py (MUSCLE3/NICE orchestration including YMMSL configs)
- added missing `__init__.py` files


waveform_editor/shape_editor/muscle3_runner.py
maarten-ic, line 1: Wouldn't it make sense to move this to `waveform_editor/shape_editor/muscle3_runner.py`? Then the panel-related stuff is inside the `waveform_editor/gui/shape_editor` folder, while the logic is just in the `waveform_editor/shape_editor` folder?

maarten-ic (commented):
Nice! Just one comment for more logical folder naming, please check if you agree (and let's discuss if needed).

waveform_editor/gui/shape_editor/xml_param/direct_param.xml
SBlokhuizen, line 1: Would it make more sense to place the `xml_param` directory in the `shape_editor` directory instead of `gui/shape_editor`?

waveform_editor/gui/shape_editor/coil_currents.py
SBlokhuizen, line 1: Can you check that every directory containing Python files also contains an `__init__.py` file? They can just be an empty file. I saw that for example `gui/shape_editor` is missing one. 

This makes sure python recognises the directory as a package:

https://docs.python.org/3/tutorial/modules.html#packages

waveform_editor/gui/shape_editor/nice_integration.py
SBlokhuizen, line 1: Most of the this class is responsible for handling the communication for the muscle actor. Would it make sense to split this and move it out of the `gui` directory?

waveform_editor/shape_editor/plasma_properties_calc.py
```diff
@@ -0,0 +1,25 @@
+"""Pure calculations for plasma profiles, with no Panel dependency."""
+
+import numpy as np
+import scipy
+
+
+def compute_profiles_from_params(ip, r0, b0, alpha, beta, gamma):
```
SBlokhuizen: `ip` and `b0` are not used here, is there a specific reason for passing them?

SBlokhuizen (commented):
Some questions, let me know if you want to discuss!

waveform_editor/shape_editor/plasma_properties_calc.py
ioan-alexandra: assumed they were gonna be used in the future when i split the plasma properties file. took them out

waveform_editor/gui/io/file_exporter.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, manager):
         Initialize the Export Dialog.
 
         Args:
-            main_gui: A reference to the main WaveformEditorGui instance.
+            manager: A reference to the WaveformEditorManager instance.
```
SBlokhuizen: This should be `IOManager` instead of `WaveformEditorManager`, right?

waveform_editor/shape_editor/muscle3_runner.py
SBlokhuizen, line 1: The ymmsl configurations and functions in this file are NICE specific. Would it make more sense to make them part of the `NiceIntegration` class, now that the panel parts have been removed from it?

SBlokhuizen (approved):

maarten-ic (approved):

PR #141, opened 2026-04-15
Run CI when new PR is created

SBlokhuizen:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,9 @@
 name: CI 
 
-on: push
+on: 
+  push:
```
maarten-ic, line 4: This will run the CI twice on PRs opened from a branch in this repository (one time for the push-trigger and another one for the PR-trigger).

You could consider restricting the push-trigger to the main and develop branches only?

maarten-ic (approved):

PR #142, opened 2026-04-17
Fix param deprecation warning

SBlokhuizen:


waveform_editor/tendencies/piecewise.py
```diff
@@ -30,6 +29,7 @@ def __init__(self, user_time=None, user_value=None, **kwargs):
             value=value,
             **kwargs,
         )
+        self.line_number = kwargs.get("line_number", 0)
```
maarten-ic: Isn't this already set by BaseTendency?

maarten-ic (commented):
Thanks! One comment:

maarten-ic (approved):

PR #143, opened 2026-04-17
update github actions to latest version

SBlokhuizen:
Update the github actions to the latest versions to get rid of the Node 20 deprecation warnings (for example [here](https://github.com/iterorganization/Waveform-Editor/actions/runs/24563102217)).


maarten-ic (approved):
LGTM

PR #145, opened 2026-04-17
restructure header navigation

### LINKED ISSUE #147: restructure the navigation
restructure the navigation on top into 2 main pages: Waveform Editor and Plasma Shape Editor. Move first 3 tabs into WE and keep the 4th one in the PSE. move the sidebar waveform creation for only WE

ioan-alexandra:
fixes #147

@maarten-ic im not sure how much custom css/js code we want for this redesign, let me know if this is overdoing. i will need to rethink some design choices 

image
image
image


maarten-ic (commented):
Hi Alexandra,

Amount of custom CSS / JS seems fine for now. Couple of comments though, let me know if anything's unclear!

1. If we don't want to show the sidebar at all when we're in the shape editor, could we instead not use it and move its content into a `Row(sidebar, waveform)` in the main content area?
2. Would be nice if the nav switcher showed which component is active:
   Image
3. Does panel provide an easy option to load the CSS from a file? That would be cleaner than having it in a string literal in the code.

waveform_editor/gui/sidebar.py
```diff
@@ -0,0 +1,36 @@
+import panel as pn
+import param
+
+
+class WaveformSidebar(param.Parameterized):
+    _open = param.Boolean(default=True)
```
maarten-ic: Let's just make this a "public" attribute, since you're also referring to this param from other classes.


```suggestion
    open = param.Boolean(default=True)
```

waveform_editor/gui/main.py
```diff
@@ -90,21 +89,64 @@ def __init__(self):
...
+        # Both pages stay in the DOM so Bokeh plots keep their layout context.
+        # Switching pages toggles visibility rather than swapping elements.
+        waveform_page = pn.Row(
+            sidebar_card,
+            tabs_card,
+            sizing_mode="stretch_both",
+            visible=is_waveform_page,
+        )
+        plasma_page = pn.Column(shape_editor, visible=is_plasma_page)
+        main_content = pn.Column(waveform_page, plasma_page, sizing_mode="stretch_both")
```
maarten-ic: This feels a bit "much" to put inside this main UI `__init__`, Perhaps you could have a `WaveformContent` class to manage this?

waveform_editor/gui/main.py
```diff
@@ -141,6 +183,23 @@ def on_tab_change(self, event):
...
+            and self.tabs.active == self.EDIT_WAVEFORMS_TAB
+            and self.editor.has_changed
+        ):
+            self.confirm_modal.show(
+                self.DISCARD_CHANGES_MESSAGE,
+                on_confirm=self.update_selection,
+                on_cancel=self.revert_to_editor,
+            )
+        else:
+            self.update_selection()
```
maarten-ic: IIRC we do this check somewhere else as well, don't we?

If so, can you please centralize the logic (so that when we change one, we won't forget about the other as well): DRY

waveform_editor/gui/sidebar.py
```diff
@@ -0,0 +1,36 @@
...
+            name="◀",
+            button_type="light",
+            width=40,
+            align=pn.bind(lambda v: "start" if v else "center", self.param._open),
+        )
+        toggle_btn.on_click(lambda _: setattr(self, "_open", not self._open))
+        self.param.watch(
+            lambda e: toggle_btn.param.update(name="◀" if e.new else "▶"),
+            "_open",
+        )
```
maarten-ic, line 16: Why not use a `pn.bind` for name as well (for consistency and fewer LoC)?

```suggestion
        toggle_btn = pn.widgets.Button(
            name=pn.bind(lambda v: "◀" if v else "▶", self.param._open),
            button_type="light",
            width=40,
            align=pn.bind(lambda v: "start" if v else "center", self.param._open),
        )
        toggle_btn.on_click(lambda _: setattr(self, "_open", not self._open))
```

waveform_editor/gui/sidebar.py
```diff
@@ -0,0 +1,36 @@
...
+
+        content = pn.Column(
+            io_manager,
+            selector,
+            confirm_modal,
+            rename_modal,
+            sizing_mode="stretch_width",
+            scroll=True,
+            visible=pn.bind(lambda v: v, self.param._open),
```
maarten-ic: I think this works as well:

```suggestion
            visible=self.param._open,
```

maarten-ic (commented):
Couple of comments, but note that I didn't run the application 😉 

Let me know if anything's unclear!

ioan-alexandra:
> Couple of comments, but note that I didn't run the application 😉

i added some screenshots so it's easier for you and sebbe to review 👍. will refactor what you mentioned

waveform_editor/gui/sidebar.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+        toggle_btn.on_click(lambda _: setattr(self, "open", not self.open))
+
+        content = pn.Column(
+            io_manager,
+            selector,
+            confirm_modal,
+            rename_modal,
+            sizing_mode="stretch_width",
+            scroll=True,
```
SBlokhuizen, line 25: Could you also change it so the width of the selector in the sidebar automatically increases if the names are too long?

Image

waveform_editor/gui/main.py
```diff
@@ -90,36 +89,49 @@ def __init__(self):
...
-            # Check if current waveform is being removed. The user already confirmed
-            # they want to remove the waveform, so we don't ask again:
-            and not self.selector.is_removing_waveform
-        ):
+    def _confirm_or_update(self, leaving_editor):
+        """Show discard-changes confirmation if needed, otherwise update selection.
+
+        leaving_editor: True when the user is navigating away from the waveform editor.
+        """
+        if leaving_editor and self.editor.has_changed:
```
SBlokhuizen, line 131: Unsaved changes in the editor are now lost when switching between waveform editor and the plasma shape editor. Could you have a look at this? 

https://github.com/user-attachments/assets/3227dc40-c4c0-4051-b8db-f2bb35b5a4ff

waveform_editor/gui/sidebar.py
ioan-alexandra: made a separate issue for this, it was already happening before this pr and its not related to the navigation

waveform_editor/gui/main.py
SBlokhuizen: Actually, can we not get rid of the warning altogether, as the unsaved waveform input is not erased anymore when switching to the Plasma Shape Editor? Then we can remove the whole `on_nav_change` method.

SBlokhuizen (approved):
Nice! LGTM

waveform_editor/gui/main.py
```diff
@@ -182,6 +196,7 @@ def load_yaml(self, yaml_content):
                 "YAML could not be loaded:<br>"
                 + self.config.load_error.replace("\n", "<br>")
             )
+        self.nav.value = WAVEFORM_EDITOR_PAGE
```
maarten-ic: Is this (still) needed?

ioan-alexandra: no, indeed. took this and another line out

maarten-ic (approved):

PR #146, opened 2026-04-28
Move infrequently used settings to separate menu

### LINKED ISSUE #152: Add NICE configurations to the Settings tab
Image

### LINKED ISSUE #156: Move rarely used settings into a dedicated settings modal

SBlokhuizen:
Replaces the old settings panel with a settings modal containing machine preset URIs, NICE settings, and display settings. 

image
image
image


waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,236 @@
...
+        """Disable machine description URI inputs if preset is selected."""
+        for inp in self._md_inputs.values():
+            inp.disabled = (
+                self.nice_settings.machine_preset != self.nice_settings.PRESET_CUSTOM
+            )
+
+    def _open_modal(self, event):
+        self.modal.show()
+
+    def _close_modal(self, event):
```
ioan-alexandra: think this is unnecessary the modal handles the closing by itself and its not called anywhere unlike the _open_modal

waveform_editor/settings.py
```diff
@@ -12,15 +12,40 @@
...
 
+class MachineDescription(param.Parameterized):
+    """Holds the URI and load state for a single machine description IDS."""
+
+    uri = param.String()
+    loaded = param.Boolean(default=False, precedence=-1)
+
+    def __init__(self, ids_name: str, **params):
+        super().__init__(**params)
+        self._ids_name = ids_name
```
ioan-alexandra: ids name is private but we access it externally in settings_modal.py and settings.py. maybe should be a public attribute or @ property?

waveform_editor/settings.py
```diff
@@ -49,18 +92,51 @@ class NiceSettings(param.Parameterized):
...
 
-    @param.depends(
-        *BASE_REQUIRED, "inv_executable", "dir_executable", "mode", watch=True
-    )
-    def check_required_params_filled(self):
-        base_ready = all(getattr(self, p) for p in self.BASE_REQUIRED)
-
-        if not base_ready:
+    @param.depends("machine_preset", watch=True)
+    def set_machine_preset(self, event=None):
```
ioan-alexandra: dont fully get how this works, event is always none, so the if after shouldnt fire ever? is this for something in the future?

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+
+import panel as pn
+import param
+from panel.viewable import Viewer
+
+from waveform_editor.gui.shape_editor.nice_plotter import NicePlotter
+from waveform_editor.gui.util import WarningIndicator
+from waveform_editor.settings import NiceSettings, settings
+
+_STYLES = [(Path(__file__).parent.parent / "styles" / "styles.css").read_text()]
```
maarten-ic: Shouldn't these be global styles that are applied on the application level?

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+def _section_label(text):
+    return pn.pane.HTML(
+        f'<p class="settings-section-label">{text}</p>',
+        stylesheets=_STYLES,
+        margin=0,
+        sizing_mode="stretch_width",
+    )
+
+
+def _card(*items):
```
maarten-ic: Naming is slightly confusing, I'd expect it to return a pn.Card, but it's actually a pn.Column.

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+                self.nice_settings.param.are_required_filled,
+            ),
+            active_icon="settings-filled",
+            description="Setting Menu",
+            size="30px",
+            on_click=lambda event: self.modal.show(),
+        )
+
+    def _build_modal(self):
+        self._md_inputs = {}
```
maarten-ic, line 76: Mention in a comment that this is for `machine description` stuff? My first association with md is markdown.

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+            scroll=True,
+        )
+
+        # --- General tab ---
+        general_content = pn.Column(
+            pn.pane.HTML(
+                '<p class="settings-placeholder">No general settings yet.</p>',
+                stylesheets=_STYLES,
+            ),
+        )
```
maarten-ic: Let's remove it then, and add it when we do have general settings?

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # --- Display tab ---
```
maarten-ic, line 101: Do we have an easy way to open the settings and jump directly to the display settings when people want to update their graph?

waveform_editor/gui/shape_editor/settings_modal.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _update_md_inputs_visibility(self, _):
+        """Disable machine description URI inputs if preset is selected."""
+        for inp in self._md_inputs.values():
+            inp.disabled = (
+                self.nice_settings.machine_preset != self.nice_settings.PRESET_CUSTOM
+            )
+
+    def __panel__(self):
+        return pn.Row(self.panel, self.modal)
```
maarten-ic: Can you just embed a modal anywhere in the UI?

waveform_editor/gui/shape_editor/shape_editor.py
```diff
@@ -65,44 +66,49 @@ def __init__(self, main_gui):
...
+        disabled_expr = (
             (
                 self.plasma_shape.param.has_shape.rx.not_()
                 & self.nice_settings.param.is_inverse_mode.rx()
             )
             | self.plasma_properties.param.has_properties.rx.not_()
             | self.nice_settings.param.are_required_filled.rx.not_()
         )
+
+        button_start = pn.widgets.Button(
```
maarten-ic, line 78: Isn't this already merged?

waveform_editor/gui/styles/styles.css
```diff
@@ -1,2 +1,57 @@
...
+    text-transform: uppercase;
+    margin: 14px 0 5px 0;
+}
+
+.settings-placeholder {
+    color: #888;
+    font-style: italic;
+    margin-top: 24px;
+    text-align: center;
+}
```
maarten-ic: Do you need placeholders?

waveform_editor/settings.py
```diff
@@ -12,15 +12,40 @@
...
+    # Preset machine description URIs
+    # TODO: Update URIs so they are from the MD database
+    ITER_PF_ACTIVE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/ITER/4/666666/3"
+    ITER_PF_PASSIVE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/ITER/4/666666/3"
+    ITER_WALL = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/ITER/4/666666/3"
+    ITER_IRON_CORE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/ITER/4/666666/3"
+    WEST_PF_ACTIVE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/west_test_dd4"
+    WEST_PF_PASSIVE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/west_test_dd4"
+    WEST_WALL = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/west_test_dd4"
+    WEST_IRON_CORE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/west_test_dd4"
```
maarten-ic: This is not portable (will only work on SDCC). We should bundle these with the application (if allowed) or try to detect if these datasets exist before offering them as preset options?

waveform_editor/settings.py
```diff
@@ -12,15 +12,40 @@
...
-        "md_pf_active",
-        "md_pf_passive",
-        "md_wall",
-        "md_iron_core",
+    PRESET_ITER = "ITER"
+    PRESET_WEST = "WEST"
+    PRESET_CUSTOM = "Custom"
+    # Preset machine description URIs
+    # TODO: Update URIs so they are from the MD database
+    ITER_PF_ACTIVE = "imas:hdf5?path=/home/ITER/blokhus/public/imasdb/ITER/4/666666/3"
```
maarten-ic: This looks a bit ugly. Would be better to 

```python
PRESET_URIS = {
    PRESET_ITER: {
        "pf_active": ...,,  # etc.
    },
},
```

That also improves the code in `set_preset()`

waveform_editor/settings.py
```diff
@@ -49,18 +92,48 @@ class NiceSettings(param.Parameterized):
     is_direct_mode = param.Boolean(precedence=-1)
     is_inverse_mode = param.Boolean(precedence=-1)
 
+    @property
+    def mds(self):
```
maarten-ic: Please write out:

```suggestion
    def machine_descriptions(self):
```

waveform_editor/settings.py
```diff
@@ -49,18 +92,48 @@ class NiceSettings(param.Parameterized):
...
+        return [self.md_pf_active, self.md_pf_passive, self.md_wall, self.md_iron_core]
+
+    def __init__(self, **params):
```
maarten-ic, line 99: Please put `__init__` as first method

waveform_editor/settings.py
maarten-ic: Does this need to be a property, or could you just create it as an attribute in `__init__` (maybe as a tuple so it cannot be edited)?

waveform_editor/settings.py
```diff
@@ -85,6 +162,13 @@ def to_dict(self):
             param_obj = self.param[p]
             if p != "name" and param_obj.precedence != -1:
                 result[p] = getattr(self, p)
+
+        if self.machine_preset == self.PRESET_CUSTOM:
+            for md in self.mds:
+                md.custom_uri = md.uri
```
maarten-ic: Doesn't feel very logical to do this as a side-effect of calling `to_dict()`.

maarten-ic (commented):
Looks nice! I have a couple of questions and comments. Overall may be good to check your docstrings (which are missing in many places).

Let me know if anything's unclear!

waveform_editor/gui/shape_editor/settings_modal.py
SBlokhuizen: I have moved the Display tab to be the first tab you see when you open the settings menu, so it requires only a single click.

waveform_editor/gui/shape_editor/settings_modal.py
SBlokhuizen: I find the holoviz documentation quite vague on this part. From what I see in the documentation and in how it is used in other examples, they usually serve the modal together with the button that triggers it, e.g. [here](https://panel.holoviz.org/reference/layouts/Modal.html#methods).

waveform_editor/gui/shape_editor/shape_editor.py
SBlokhuizen: This just adds a reactive tooltip to the run button.

waveform_editor/gui/shape_editor/settings_modal.py
ioan-alexandra: yes, i've moved now the path in the utils and will be importing it from there from now on. unfortunately i'll still need to declare `stylesheets=STYLES` on some components because panel is annoying and rendering stuff inside the browsers shadow DOM which ignores the parent css but at least the path will only be declared once

maarten-ic: Ah right 😞 , and what if you'd use the `:root` selector? https://panel.holoviz.org/how_to/styling/design_variables.html#:~:text=When%20defining%20design%20variable%20overrides%20globally%20use%20the%20%3Aroot%20CSS%20selector

I haven't tested it, but it would be nice if we wouldn't need to repeatedly add a stylesheet in the application.

waveform_editor/settings.py
```diff
@@ -92,48 +96,76 @@ class NiceSettings(param.Parameterized):
...
+                logger.warning(
+                    "Machine Description Preset '%s' could not be loaded: %s",
+                    preset_name,
+                    err,
+                )
+                continue
+
+            available[preset_name] = preset
+
+        self.PRESET_URIS = available
```
maarten-ic: I wouldn't shadow a class attribute that appears constant (all-caps name)

maarten-ic (approved):

PR #157, opened 2026-04-30
use radiobuttons instead of checkboxes for nice mode and update stop/run buttons

### LINKED ISSUE #155: Use toggle for direct-inverse option
Image

ioan-alexandra:
fixes #155

image


waveform_editor/gui/settings.py
```diff
@@ -7,6 +7,40 @@
...
+
+    Returns:
+        A Panel Row with a Switch and a reactive mode label.
+    """
+    toggle = pn.widgets.Switch(
+        value=nice_settings.is_direct_mode,
+        margin=0,
+        stylesheets=[":host { transform: scale(1.5); transform-origin: left center; }"],
+    )
+    nice_settings.param.watch(
```
SBlokhuizen: To me this makes it seem like NICE Inverse mode disables something, while direct mode enables something. Do you think this makes more sense than a radio button like before?

Image

Image

waveform_editor/gui/settings.py
```diff
@@ -7,6 +7,22 @@
...
+    """Create a segmented control for switching between NICE modes."""
+    widget = pn.widgets.RadioButtonGroup(
+        options=[NiceSettings.DIRECT_MODE, NiceSettings.INVERSE_MODE],
+        value=nice_settings.mode,
+        button_type="default",
+        button_style="outline",
+        **kwargs,
+    )
+    nice_settings.param.watch(lambda e: setattr(widget, "value", e.new), "mode")
+    widget.param.watch(lambda e: setattr(nice_settings, "mode", e.new), "value")
```
SBlokhuizen: Instead of having two watchers, can you not use a link here  with `bidirectional=True`?

https://panel.holoviz.org/how_to/links/links.html

SBlokhuizen (approved):
LGTM!

maarten-ic (approved):

PR #160, opened 2026-05-05
properties tab select between manual and eq ids

### LINKED ISSUE #153: In properties tab allow user to select between manual and Eq IDS for the input
Image

ioan-alexandra:
fixes #153

image

### DIFF
--- a/waveform_editor/gui/shape_editor/plasma_properties.py
+++ b/waveform_editor/gui/shape_editor/plasma_properties.py
@@ -5,6 +5,7 @@ import panel as pn
 import param
 from panel.viewable import Viewer
 
+from waveform_editor.gui.util import FormattedEditableFloatSlider
 from waveform_editor.shape_editor.plasma_properties_calc import (
     compute_profiles_from_params,
 )
@@ -30,42 +31,29 @@ class PropertyInput(Viewer):
         self.value = default_value
         self._label = label
 
-        self._mode_toggle = pn.widgets.RadioButtonGroup(
-            options=[MANUAL, EQ_IDS],
-            value=self.mode,
+        self._mode_toggle = pn.widgets.RadioButtonGroup.from_param(
+            self.param.mode,
             button_style="outline",
             button_type="primary",
             stylesheets=[_CARD_CSS],
         )
-        self._mode_toggle.param.watch(self._on_mode_change, "value")
-
-        self._value_input = pn.widgets.FloatInput(
-            value=self.value,
+        self._value_input = pn.widgets.FloatInput.from_param(
+            self.param.value,
             step=step,
             sizing_mode="stretch_width",
             margin=(4, 0, 0, 0),
         )
-        self._value_input.param.watch(self._on_value_change, "value")
-
         self._uri_input = pn.widgets.TextInput.from_param(
             self.param.ids_uri,
-            name="",
-            placeholder="IDS URI",
+            name="IDS URI",
             sizing_mode="stretch_width",
         )
         self._time_input = pn.widgets.FloatInput.from_param(
             self.param.ids_time, name="Time [s]", width=100
         )
-        self.param.watch(
-            lambda *_: self.param.trigger("changed"), ["ids_uri", "ids_time"]
-        )
 
-    def _on_mode_change(self, event):
-        self.mode = event.new
-        self.param.trigger("changed")
-
-    def _on_value_change(self, event):
-        self.value = event.new
+    @param.depends("mode", "value", "ids_uri", "ids_time", watch=True)
+    def _on_change(self):
         self.param.trigger("changed")
 
     def __panel__(self):
@@ -80,7 +68,7 @@ class PropertyInput(Viewer):
         return pn.Column(
             header,
             pn.Column(self._value_input, visible=is_manual, margin=(4, 0, 0, 0)),
-            pn.Column(
+            pn.Row(
                 self._uri_input, self._time_input, visible=is_ids, margin=(4, 0, 0, 0)
             ),
             css_classes=["property-card"],
@@ -99,53 +87,43 @@ class PlasmaProfiles(Viewer):
     """
 
     mode = param.ObjectSelector(default=PARAMETRIC, objects=[PARAMETRIC, EQ_IDS])
-    alpha = param.Number(default=0.5, step=0.01)
-    beta = param.Number(default=0.5, step=0.01)
-    gamma = param.Number(default=1.0, step=0.01)
+    alpha = param.Number(default=0.5, softbounds=[0.5, 2], step=0.01)
+    beta = param.Number(default=0.5, softbounds=[0.5, 2], step=0.01)
+    gamma = param.Number(default=1.0, softbounds=[0.5, 2], step=0.01)
     ids_uri = param.String(default="")
     ids_time = param.Number(default=0.0)
     changed = param.Event()
+    profiles_plot = param.Parameter(default=None)
 
     def __init__(self, **params):
         super().__init__(**params)
 
-        self._mode_toggle = pn.widgets.RadioButtonGroup(
-            options=[PARAMETRIC, EQ_IDS],
-            value=self.mode,
+        self._mode_toggle = pn.widgets.RadioButtonGroup.from_param(
+            self.param.mode,
             button_style="outline",
             button_type="primary",
             stylesheets=[_CARD_CSS],
         )
-        self._mode_toggle.param.watch(self._on_mode_change, "value")
-
-        self._alpha_input = pn.widgets.FloatInput.from_param(
-            self.param.alpha, name="Alpha", sizing_mode="stretch_width"
+        self._alpha_input = FormattedEditableFloatSlider.from_param(
+            self.param.alpha, name="Alpha", margin=0
         )
-        self._beta_input = pn.widgets.FloatInput.from_param(
-            self.param.beta, name="Beta", sizing_mode="stretch_width"
+        self._beta_input = FormattedEditableFloatSlider.from_param(
+            self.param.beta, name="Beta", margin=0
         )
-        self._gamma_input = pn.widgets.FloatInput.from_param(
-            self.param.gamma, name="Gamma", sizing_mode="stretch_width"
+        self._gamma_input = FormattedEditableFloatSlider.from_param(
+            self.param.gamma, name="Gamma", margin=0
         )
-        self.param.watch(
-            lambda *_: self.param.trigger("changed"), ["alpha", "beta", "gamma"]
-        )
-
         self._uri_input = pn.widgets.TextInput.from_param(
             self.param.ids_uri,
-            name="",
-            placeholder="IDS URI",
+            name="IDS URI",
             sizing_mode="stretch_width",
         )
         self._time_input = pn.widgets.FloatInput.from_param(
             self.param.ids_time, name="Time [s]", width=100
         )
-        self.param.watch(
-            lambda *_: self.param.trigger("changed"), ["ids_uri", "ids_time"]
-        )
 
-    def _on_mode_change(self, event):
-        self.mode = event.new
+    @param.depends("mode", "alpha", "beta", "gamma", "ids_uri", "ids_time", watch=True)
+    def _on_change(self):
         self.param.trigger("changed")
 
     def __panel__(self):
@@ -164,6 +142,7 @@ class PlasmaProfiles(Viewer):
                 self._beta_input,
                 self._gamma_input,
                 visible=is_parametric,
+                align="end",
                 margin=(4, 0, 0, 0),
             ),
             pn.Column(
@@ -172,6 +151,7 @@ class PlasmaProfiles(Viewer):
                 visible=is_ids,
                 margin=(4, 0, 0, 0),
             ),
+            pn.bind(lambda p: p, self.param.profiles_plot),
             css_classes=["property-card"],
             stylesheets=[_CARD_CSS],
             max_width=600,
@@ -185,9 +165,17 @@ class PlasmaProperties(Viewer):
     )
     has_properties = param.Boolean(doc="Whether the plasma properties are loaded.")
 
+    @property
+    def profiles_plot(self):
+        return self._profiles.profiles_plot
+
+    @profiles_plot.setter
+    def profiles_plot(self, value):
+        self._profiles.profiles_plot = value
+
     def __init__(self):
         super().__init__()
-        self._ip = PropertyInput("Plasma current [A]", default_value=-1.5e7)
+        self._ip = PropertyInput("Plasma current [A]", default_value=-1.5e7, step=1e6)
         self._r0 = PropertyInput("Reference major radius [m]", default_value=6.2)
         self._b0 = PropertyInput("Toroidal magnetic field [T]", default_value=-5.3)
         self._profiles = PlasmaProfiles()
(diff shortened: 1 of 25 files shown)


waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,278 @@
...
+            self._mode_toggle,
+            sizing_mode="stretch_width",
+            margin=0,
+        )
+        return pn.Column(
+            header,
+            pn.Column(self._value_input, visible=is_manual, margin=(4, 0, 0, 0)),
+            pn.Column(
+                self._uri_input, self._time_input, visible=is_ids, margin=(4, 0, 0, 0)
+            ),
```
SBlokhuizen: Do you think it makes sense to put the uri and time input in a row to save vertical space?

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,278 @@
...
+
+        self._alpha_input = pn.widgets.FloatInput.from_param(
+            self.param.alpha, name="Alpha", sizing_mode="stretch_width"
+        )
+        self._beta_input = pn.widgets.FloatInput.from_param(
+            self.param.beta, name="Beta", sizing_mode="stretch_width"
+        )
+        self._gamma_input = pn.widgets.FloatInput.from_param(
+            self.param.gamma, name="Gamma", sizing_mode="stretch_width"
+        )
```
SBlokhuizen, line 156: I think for the alpha, beta, and gamma inputs it does make sense to make them EditableFloatSliders instead.

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,278 @@
...
         super().__init__()
-        self.indicator = WarningIndicator(visible=self.param.has_properties.rx.not_())
-        self.radio_box = pn.widgets.RadioBoxGroup.from_param(
-            self.param.input_mode, inline=True, margin=(15, 20, 0, 20)
-        )
-        self.panel = pn.Column(self.radio_box, self._panel_property_options)
+        self._ip = PropertyInput("Plasma current [A]", default_value=-1.5e7)
+        self._r0 = PropertyInput("Reference major radius [m]", default_value=6.2)
+        self._b0 = PropertyInput("Toroidal magnetic field [T]", default_value=-5.3)
+        self._profiles = PlasmaProfiles()
```
SBlokhuizen, line 349: Can you move the profiles plot into the same card as the `Profile source`?

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,278 @@
...
+        self._label = label
+
+        self._mode_toggle = pn.widgets.RadioButtonGroup(
+            options=[MANUAL, EQ_IDS],
+            value=self.mode,
+            button_style="outline",
+            button_type="primary",
+            stylesheets=[_CARD_CSS],
+        )
+        self._mode_toggle.param.watch(self._on_mode_change, "value")
```
SBlokhuizen: You could also use `@param.depends` for these cases instead of a watcher:

https://panel.holoviz.org/how_to/param/dependencies.html

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,278 @@
...
     has_properties = param.Boolean(doc="Whether the plasma properties are loaded.")
 
     def __init__(self):
         super().__init__()
-        self.indicator = WarningIndicator(visible=self.param.has_properties.rx.not_())
-        self.radio_box = pn.widgets.RadioBoxGroup.from_param(
-            self.param.input_mode, inline=True, margin=(15, 20, 0, 20)
-        )
-        self.panel = pn.Column(self.radio_box, self._panel_property_options)
+        self._ip = PropertyInput("Plasma current [A]", default_value=-1.5e7)
```
SBlokhuizen: Nominal Ip is around 15MA, so let's increase the step size here to `1e6`.

waveform_editor/gui/shape_editor/plasma_properties.py
ioan-alexandra: can do

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,266 @@
...
+            header,
+            pn.Column(self._value_input, visible=is_manual, margin=(4, 0, 0, 0)),
+            pn.Row(
+                self._uri_input, self._time_input, visible=is_ids, margin=(4, 0, 0, 0)
+            ),
+            css_classes=["property-card"],
+            stylesheets=[_CARD_CSS],
+            max_width=600,
+            margin=(0, 0, 8, 0),
+        )
```
SBlokhuizen, line 118: It would be nice to show what value is loaded if you load from an IDS.

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,283 @@
...
-            beta=self.properties_params.beta,
-            gamma=self.properties_params.gamma,
-        )
-        self.has_properties = True
-
-    def _load_properties_from_ids(self):
-        """Load plasma properties from IDS equilibrium input."""
-        if not self.input.uri:
-            self.has_properties = False
+    def _load_profiles_from_ids(self, uri: str, time: float):
```
SBlokhuizen: Can you bring back / update the docstrings? Same for the other methods without docstrings.

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,283 @@
...
     has_properties = param.Boolean(doc="Whether the plasma properties are loaded.")
 
+    @property
+    def profiles_plot(self):
+        return self._profiles.profiles_plot
+
+    @profiles_plot.setter
+    def profiles_plot(self, value):
+        self._profiles.profiles_plot = value
+
```
SBlokhuizen, line 333: Doesn't it make more sense to move the profile plotting logic to the `PlasmaProfiles` class, instead of having to pass it through from the `NicePlotter`?

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,223 @@
...
-    FormattedEditableFloatSlider,
-    WarningIndicator,
-)
+from waveform_editor.gui.util import FormattedEditableFloatSlider
 from waveform_editor.shape_editor.plasma_properties_calc import (
     compute_profiles_from_params,
 )
 
+MANUAL = "Manual"
+EQ_IDS = "Eq IDS"
```
maarten-ic: I'd prefer to spell this out, if we have sufficient screenspace

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,223 @@
...
-        default=-5.3, softbounds=[-10, 10], label="Toroidal field at R0 [T]"
-    )
-    alpha = param.Number(default=0.5, softbounds=[0.5, 2], step=0.01, label="Alpha")
-    beta = param.Number(default=0.5, softbounds=[0.5, 2], step=0.01, label="Beta")
-    gamma = param.Number(default=1.0, softbounds=[0.5, 2], step=0.01, label="Gamma")
+
+class PropertyInput(Viewer):
+    """A single scalar plasma property with its own Manual / Eq IDS mode toggle."""
+
+    mode = param.ObjectSelector(default=MANUAL, objects=[MANUAL, EQ_IDS])
```
maarten-ic: `ObjectSelector` is deprecated, can you use `Selector` instead?

https://github.com/holoviz/param/blob/b8dcf8d62582f8d11acccc6f80ea14e38620fd83/param/parameters.py#L2745-L2749

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,223 @@
...
+            pn.pane.HTML(self._label, sizing_mode="stretch_width", margin=(5, 0)),
+            self._mode_toggle,
+            sizing_mode="stretch_width",
+            margin=0,
+        )
+        return pn.Column(
+            header,
+            pn.Column(self._value_input, visible=is_manual, margin=(4, 0, 0, 0)),
+            pn.Column(
+                pn.Row(self._uri_input, self._time_input, margin=(4, 0, 0, 0)),
```
maarten-ic: Do you need these columns when there's only a single item inside it?

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,223 @@
...
+
+class PlasmaProfiles(Viewer):
+    """Widget for selecting the plasma profile source: parametric or from IDS.
+
+    In Parametric mode, dpressure_dpsi and f_df_dpsi are computed analytically
+    from alpha/beta/gamma shape parameters. In Eq IDS mode, they are read
+    directly from an equilibrium IDS file.
+    """
+
+    mode = param.ObjectSelector(default=PARAMETRIC, objects=[PARAMETRIC, EQ_IDS])
```
maarten-ic: idem

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -58,83 +226,106 @@ class PlasmaProperties(Viewer):
...
+        if not prop.ids_uri:
+            return None
+        try:
+            with imas.DBEntry(prop.ids_uri, "r") as entry:
+                eq = entry.get_slice(
+                    "equilibrium", prop.ids_time, imas.ids_defs.CLOSEST_INTERP
+                )
+            value = extractor(eq)
+            prop.loaded_value = float(value)
+            return value
```
maarten-ic: Let's use lazy loading and string path instead of an extractor function?

```suggestion
            with imas.DBEntry(prop.ids_uri, "r") as entry:
                eq = entry.get_slice(
                    "equilibrium", prop.ids_time, imas.ids_defs.CLOSEST_INTERP, lazy=True
                )
                prop.loaded_value = value = float(eq[path])
            return value
```

Can be used as 
```python
self.b0 = self._load_scalar(self._b0, "vacuum_toroidal_field.b0[0]")
```

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -58,83 +226,106 @@ class PlasmaProperties(Viewer):
...
         self.psi_norm = None
         self.ip = None
         self.r0 = None
         self.b0 = None
 
-    @param.depends(
-        "properties_params.param", "input.param", "input_mode", watch=True, on_init=True
-    )
+        for widget in [self._ip, self._r0, self._b0, self._profiles]:
+            widget.param.watch(lambda *_: self._load_plasma_properties(), "changed")
```
maarten-ic: It's better to add an unused `_event` argument to `self._load_plasma_properties` so you can just do:

```suggestion
            widget.param.watch(self._load_plasma_properties, "changed")
```

That avoids creating 4 anonymous lambda functions, and it's slightly more readable here

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -58,83 +226,106 @@ class PlasmaProperties(Viewer):
...
-        if self.input_mode == self.EQUILIBRIUM_INPUT:
-            self._load_properties_from_ids()
-        elif self.input_mode == self.MANUAL_INPUT:
-            self._load_properties_from_params()
+        Resolves each scalar independently, then either reads profile functions from an
+        IDS or computes them parametrically. Triggers `profile_updated` when done.
+        """
+        self.ip = self._load_scalar(
+            self._ip, lambda eq: eq.time_slice[0].global_quantities.ip
+        )
```
maarten-ic: Could we do this in `PropertyInput` instead? Now we need to load and reload ip even when the user changes something else entirely

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -58,83 +226,106 @@ class PlasmaProperties(Viewer):
...
+        self.r0 = self._load_scalar(self._r0, lambda eq: eq.vacuum_toroidal_field.r0)
+        self.b0 = self._load_scalar(self._b0, lambda eq: eq.vacuum_toroidal_field.b0[0])
+
+        if self._profiles.mode == EQ_IDS:
+            self._load_profiles_from_ids(
```
maarten-ic: Idem, let's move to `PlasmaProfiles` instead?

maarten-ic (commented):
UI improvement looks good! Some small comments below, and let's review the architecture: I think the code will be a lot cleaner if the data loading is done "locally" on the viewer classes.

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -226,9 +314,22 @@ class PlasmaProperties(Viewer):
...
+        self._r0 = PropertyInput(
+            "Reference major radius [m]",
+            default_value=6.2,
+            ids_path="vacuum_toroidal_field.r0",
+        )
+        self._b0 = PropertyInput(
+            "Toroidal magnetic field [T]",
+            default_value=-5.3,
+            ids_path="vacuum_toroidal_field.b0[0]",
+        )
```
SBlokhuizen, line 348: Could you test out loading these values from an equilibrium IDS to make sure it works? It currently crashes when trying to load the IDS path. I think you need to use slashes instead of dots for the path.

SBlokhuizen (changes requested):

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,311 @@
...
+        default_value:
+            Initial value used in Manual mode.
+        ids_path:
+            Dot-path into the equilibrium IDS used in Eq IDS mode, e.g.
+            ``"vacuum_toroidal_field.r0"``.
+        step:
+            Step size for the numeric input widget.
+        """
+        super().__init__(**params)
+        self._ids_path = ids_path
```
mikesndrs: I imagine many people will use a single source IDS and adding the same URI over and over feels cumbersome. Probably good to have some way of reusing the equilibrium IDS path that is already present somewhere. 
Maybe double check if this is necessary/wanted?

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,55 +1,311 @@
...
+        self.value = default_value
+        self._label = label
```
mikesndrs, line 61: probably not this PR but error messages disappear before I can fully read them

waveform_editor/gui/shape_editor/plasma_properties.py
ioan-alexandra: fixed

waveform_editor/gui/shape_editor/plasma_properties.py
ioan-alexandra: fixed in separate pr

waveform_editor/gui/shape_editor/plasma_properties.py
```diff
@@ -1,140 +1,431 @@
...
+            header,
+            self._value_input,
+            css_classes=["property-card"],
+            stylesheets=[_CARD_CSS],
+            max_width=600,
+            margin=(0, 0, 8, 0),
+        )
+
+
+class PlasmaProfiles(Viewer):
```
SBlokhuizen, line 121: Image

When you load the profiles from an IDS, it still shows the alpha, beta and gamma parameterisation values even though it is not using a parameterisation, which is a bit confusing. Perhaps we can hide them when loading from an IDS?

mikesndrs (approved):

SBlokhuizen (approved):

PR #165, opened 2026-05-20
Show NICE metrics 

SBlokhuizen:
Shows the following parameters after a NICE run:

image


ioan-alexandra (commented):
not vibing with the metrics being shown empty before the simulation is run and how its pushing the graph down. and its clipping into the menu. need to think a bit about where to place these

Image

ioan-alexandra (approved):

PR #168, opened 2026-06-22
Add whole-trace / overlay mode to the MUSCLE3 waveform actor

DaanVanVugt:
Previously the actor exported a single time slice per reuse at the incoming message timestamp. Now, when the input port is named '<ids>_in' and the F_INIT message carries that IDS, the actor reads its /time and exports the waveforms interpolated onto every time present, in one message -- falling back to the message timestamp otherwise.

The incoming IDS is also used as the export base (new `base_idss` on ConfigurationExporter, sharing the machine-description resolution path), so its existing data is preserved and the waveforms are overlaid in place. This lets the actor sit inline in a pipeline and augment an IDS passing through -- e.g. adding Ip to an equilibrium on its way to a solver -- rather than emitting a fresh IDS.

Both modes are covered by test_muscle3.

### DIFF
--- a/waveform_editor/export/exporter.py
+++ b/waveform_editor/export/exporter.py
@@ -13,10 +13,13 @@ logger = logging.getLogger(__name__)
 
 
 class ConfigurationExporter:
-    def __init__(self, config, times, progress=None):
+    def __init__(self, config, times, progress=None, base_idss=None):
         self.config = config
         self.times = times
         self.progress = progress
+        # {ids_name: IDS} bases to overlay the waveforms onto in place (preserving their
+        # other data), taking precedence over the machine description / an empty IDS.
+        self.base_idss = base_idss or {}
         self.total_progress = None
         self.current_progress = None
         # We assume that all DD times are in seconds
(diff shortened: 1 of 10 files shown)


SBlokhuizen (commented):
Code looks good to me, could you update [the documentation](https://waveform-editor.readthedocs.io/en/latest/muscle3.html) as well?

waveform_editor/export/exporter.py
```diff
@@ -70,15 +73,7 @@ def _generate_idss(self, factory):
...
-            if md:
-                with imas.DBEntry(md, "r") as entry_md:
-                    orig_ids = entry_md.get(ids_name, autoconvert=False)
-                    ids = imas.convert_ids(orig_ids, self.config.globals.dd_version)
-            else:
-                ids = factory.new(ids_name)
+            ids = self._base_ids(ids_name, factory)
             # TODO: currently only IDSs with homogeneous time mode are supported
             ids.ids_properties.homogeneous_time = (
                 imas.ids_defs.IDS_TIME_MODE_HOMOGENEOUS
```
SBlokhuizen, line 85: Should we raise an error if a non-homogeneous IDS is passed instead of silently converting to a homogeneous IDS?

DaanVanVugt: I would raise only if /time is not present, and log a warning if inhomogeneous time mode is used

SBlokhuizen (approved):

waveform_editor/export/exporter.py
maarten-ic: We should raise an error when trying to amend an inhomogeneous IDS, otherwise we probably create incorrect data.

It's better to explicitly fail for an unsupported use case, logged warnings are typically ignored

PR #170, opened 2026-06-23
Load values from external IDSes

DaanVanVugt:
After #168 
Also allows loading 2D values, but then cannot be combined with other tendencies in the same variable.
Also allows wildcards in the IDS path.

### DIFF
--- a/docs/source/examples/imports.yaml
+++ b/docs/source/examples/imports.yaml
@@ -47,8 +47,3 @@ Plasma current:
   # source is runtime data rather than a file.
   equilibrium/*:
     - {ref: live_eq}
-
-# A bare `*` overlays every IDS a source provides (sources stack, applied in order);
-# more specific imports above still win. Omitted here as it would overlay all IDSs:
-#   '*':
-#     - {ref: machine}
--- a/docs/source/examples/imports.yaml
+++ b/docs/source/examples/imports.yaml
@@ -1,0 +1,49 @@
+# Reference example for the `imports` mechanism. The URIs are illustrative.
+globals:
+  dd_version: 4.0.0
+  imports:
+    machine:  imas:hdf5?path=/path/to/machine_description    # static (no /time)
+    scenario: imas:hdf5?path=/path/to/scenario_run           # time-dependent
+    live_eq:  {port: equilibrium_in}                         # received at run time
+
+ec_launchers:
+  # Whole-IDS import = machine-description overlay: copied in first (as a base),
+  # then the leaves below override individual nodes.
+  ec_launchers/*:
+    - {ref: machine}
+  ec_launchers/beam(1)/phase/angle:
+    - {type: linear, to: 1.65, duration: 100}
+
+Heating:
+  # Scalar (0D) import, resampled onto the export /time with linear interpolation.
+  core_sources/source(1)/global_quantities/total_ion_power:
+    - {ref: scenario, interp: linear}
+
+  # Static value (not a time series) -- names the source.
+  core_sources/source(1)/identifier/name:
+    - {value: ec}
+
+  # Trailing /* -- mirror every filled leaf of the subtree from the entry.
+  core_sources/source(1)/profiles_1d/*:
+    - {ref: scenario}
+
+  # Index wildcard -- import the leaf for every source.
+  core_sources/source(*)/global_quantities/total_ion_power:
+    - {ref: scenario}
+
+  # Multiple index wildcards -- every (source, ion) combination.
+  core_sources/source(*)/profiles_1d/ion(*)/z_ion:
+    - {ref: scenario}
+
+Plasma current:
+  # Composite (0D only): an analytic segment, then an imported segment. Each fills its
+  # [start, end] window.
+  equilibrium/time_slice/global_quantities/ip:
+    - {type: constant, value: -1.0e6, duration: 1}
+    - {ref: scenario, duration: 1, time_offset: 0.5}
+
+  # Overlay onto a live IDS received over a MUSCLE3 port (the actor's equilibrium_in),
+  # then override a leaf. Same shape as the machine-description overlay above, but the
+  # source is runtime data rather than a file.
+  equilibrium/*:
+    - {ref: live_eq}
--- a/docs/source/examples/imports.yaml
+++ b/docs/source/examples/imports.yaml
@@ -47,3 +47,8 @@ Plasma current:
   # source is runtime data rather than a file.
   equilibrium/*:
     - {ref: live_eq}
+
+# A bare `*` overlays every IDS a source provides (sources stack, applied in order);
+# more specific imports above still win. Omitted here as it would overlay all IDSs:
+#   '*':
+#     - {ref: machine}
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -211,18 +211,18 @@ Parameters
     core_sources/source(1)/profiles_1d/electrons/energy:
       - {ref: scenario, interp: linear}
 
-Wildcards expand against the source: a trailing ``*`` imports a whole subtree (``<ids>/*`` copies a whole IDS), and a ``(*)`` index wildcard iterates every element of an array of structure -- several may be combined, e.g. every ion of every source. A bare ``*`` imports the whole entry: every IDS the source provides (a URI is scanned for filled IDSs; a ``{port: ...}`` source contributes the IDS it received), loading a machine description in one line. An overlay may list several sources, applied in order:
+Wildcards expand against the source: a trailing ``*`` imports a whole subtree (``<ids>/*`` copies a whole IDS), and a ``(*)`` index wildcard iterates every element of an array of structure -- several may be combined, e.g. every ion of every source. An overlay may list several sources, applied in order:
 
 .. code-block:: yaml
 
     core_sources/source(*)/profiles_1d/ion(*)/z_ion:
       - {ref: scenario}
 
-    '*':                  # whole-entry overlay; sources stack
+    ec_launchers/*:        # whole-IDS overlay; sources stack
       - {ref: machine}
       - {ref: scenario}
 
-**Precedence.** Where imports or explicit waveforms write the same node, the most specific wins regardless of order: ``*`` < ``<ids>/*`` < subtree ``.../*`` < explicit leaf. Equal specificity falls back to listing order (last wins).
+**Precedence.** Where imports or explicit waveforms write the same node, the most specific wins regardless of order: ``<ids>/*`` < subtree ``.../*`` < explicit leaf. Equal specificity falls back to listing order (last wins).
 
 Non-0D imports (a profile, a wildcard subtree) own the whole waveform. Only **0D (scalar)** imports combine with analytic segments, each filling its ``[start, end]`` window:
 
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -192,6 +192,42 @@ Parameters
 .. warning::
     This tendency does **not** accept the common ``start``, ``duration``, or ``end`` parameters. These are derived directly from the required ``time`` list.
 
+.. _import-tendency:
+
+Import
+======
+
+Takes its values from an external entry declared in :ref:`globals.imports <global_properties>` instead of an analytic shape. By default the same DD path the waveform sits at is read from the import and resampled onto the export time base.
+
+*Type:* ``import`` (inferred when ``ref`` is present)
+
+Parameters
+----------
+*   ``ref``: Name of the entry in ``globals.imports`` to read from.
+*   ``path``: DD path to read from the import. Defaults to the waveform's own path.
+*   ``time_offset``: Offset added to the export time when sampling the import. Defaults to ``0``.
+*   ``interp``: Resampling mode onto the export time base: ``closest`` (default), ``linear`` or ``previous``.
+
+.. code-block:: yaml
+
+    core_sources/source(1)/profiles_1d/electrons/energy:
+      - {ref: scenario, interp: linear}
+
+Wildcards expand against the source: a trailing ``*`` imports every filled leaf of that subtree (e.g. ``core_sources/source(1)/profiles_1d/*``), an ``<ids>/*`` import copies a whole IDS and acts as an overlay base, and a ``(*)`` index wildcard imports the leaf for every element of that array of structure. Several ``(*)`` may be combined for higher-dimensional imports, e.g. every ion of every source:
+
+.. code-block:: yaml
+
+    core_sources/source(*)/profiles_1d/ion(*)/z_ion:
+      - {ref: scenario}
+
+Non-0D imports (a value per radial point, a wildcard subtree) own the whole waveform. Only **0D (scalar)** imports may be combined with analytic segments, each filling its ``[start, end]`` window:
+
+.. code-block:: yaml
+
+    equilibrium/time_slice/global_quantities/ip:
+      - {type: constant, value: -1.0, duration: 1}   # analytic on [0, 1] s
+      - {ref: scenario, duration: 1}                 # imported on [1, 2] s
+
 Periodic Tendencies
 ===================
 
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -8,6 +8,10 @@ This document describes the different types of tendencies available in the Wavef
 
 Each tendency defines the behavior of the signal over a specific time interval. You can chain multiple tendencies together to create complex waveforms.
 
+The ``type`` may be omitted, in which case it is inferred from the entry's keys: ``to`` implies a :ref:`Linear <available-tendencies>` tendency, ``time`` a :ref:`Piecewise Linear <piecewise-linear-tendency>` tendency, and ``value`` a :ref:`Constant <available-tendencies>` tendency. Anything else falls back to ``linear`` -- so a linear tendency does *not* require ``to``; a ``from``-only, ``rate``-only, or bare segment is still linear and takes its endpoints from its neighbours.
+
+A segment whose distinguishing key is absent must therefore name its ``type`` explicitly: the periodic shapes (sine, sawtooth, triangle, square), a :ref:`Smooth <available-tendencies>` tendency, and a value-less :ref:`Constant <available-tendencies>` (a bare ``{duration: ...}`` is read as a linear ramp, not a held constant).
+
 Common Time Parameters
 ======================
 
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -8,9 +8,7 @@ This document describes the different types of tendencies available in the Wavef
 
 Each tendency defines the behavior of the signal over a specific time interval. You can chain multiple tendencies together to create complex waveforms.
 
-The ``type`` may be omitted, in which case it is inferred from the entry's keys: ``to`` implies a :ref:`Linear <available-tendencies>` tendency, ``time`` a :ref:`Piecewise Linear <piecewise-linear-tendency>` tendency, and ``value`` a :ref:`Constant <available-tendencies>` tendency. Anything else falls back to ``linear`` -- so a linear tendency does *not* require ``to``; a ``from``-only, ``rate``-only, or bare segment is still linear and takes its endpoints from its neighbours.
-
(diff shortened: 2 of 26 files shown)


waveform_editor/tendencies/reference.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
+import param
+
+from waveform_editor.tendencies.base import BaseTendency
+
+
+class ReferenceTendency(BaseTendency):
```
SBlokhuizen: This doesn't really feel like a tendency to me. To me a tendency should be a self-contained value generator, but this is pushing the work to the exporter. Would it make more sense to have a separate (derived) classes for the reference waveforms and static variables?

waveform_editor/waveform.py
```diff
@@ -186,6 +188,23 @@ def update_annotations(self, event=None):
...
+    def _infer_tendency_type(entry):
+        """Infer the tendency type from an entry's keys when ``type`` is omitted.
+
+        Unambiguous keys map directly; periodic shapes share keys (period/amplitude) so
+        they still need an explicit ``type``. Falls back to ``linear``.
+        """
+        for key, tendency_type in (
+            ("user_ref", "reference"),
+            ("user_to", "linear"),
+            ("user_points", "piecewise"),
```
SBlokhuizen: `points` is not an argument of a piecewise tendency, instead it stores the points in `time` and `value` arrays: https://waveform-editor.readthedocs.io/en/latest/tendencies.html#piecewise-linear-tendency.

waveform_editor/configuration.py
```diff
@@ -27,6 +27,9 @@ def __init__(self):
         # of waveforms
         self.waveform_map = {}
         self.globals = YamlGlobals()
+        # Top-level `references` section: named external data entries (name -> imas URI)
+        # read by `reference` tendencies.
+        self.references = {}
```
SBlokhuizen: Wouldn't it make more sense to let the references be part of the `YamlGlobals` class? Also users cannot change the references from the UI currently. Perhaps add an input field [here](https://waveform-editor.readthedocs.io/en/latest/gui.html#editing-global-properties)?

tests/test_exporter.py
```diff
@@ -709,6 +709,144 @@ def test_example_yaml(tmp_path):
     assert np.all(nbi.unit[0].energy.data == values)
 
 
+@pytest.fixture
+def core_sources_ext_uri(tmp_path):
+    """An external, time-dependent core_sources to import from (unnamed source)."""
```
SBlokhuizen: What is (unnamed source)?

tests/test_exporter.py
```diff
@@ -709,6 +709,144 @@ def test_example_yaml(tmp_path):
...
+globals:
+  dd_version: 4.0.0
+Heating:
+  core_sources/source(1)/profiles_1d/electrons/energy:
+    - {{type: reference, ref: ext}}
+  core_sources/source(1)/identifier/name:
+    - {{value: ec}}
+"""
+    )
+    times = np.array([0.0, 1.0, 2.0])
```
SBlokhuizen, line 792: What would happen with the external waveforms if the timebase wouldn't exactly match the one used in your fixture? Should we support different interpolation modes?

tests/test_exporter.py
```diff
@@ -709,6 +709,144 @@ def test_example_yaml(tmp_path):
...
+    """A `*` after a prefix imports every filled leaf of that subtree from the ref."""
+    config = WaveformConfiguration()
+    config.load_yaml(
+        f"""
+references:
+  ext: "{core_sources_ext_uri}"
+globals:
+  dd_version: 4.0.0
+Heating:
+  core_sources/source(1)/profiles_1d/*:
```
SBlokhuizen, line 855: Would be nice to support wildcards on the indices as well. E.g. If I want the ion charge of the first ion for all sources:  

`source(*)/profiles_1d/ion(0)/z_io`

tests/test_exporter.py
```diff
@@ -709,6 +709,144 @@ def test_example_yaml(tmp_path):
     assert np.all(nbi.unit[0].energy.data == values)
 
 
+@pytest.fixture
```
SBlokhuizen, line 756: General comment: could you update the docs?

waveform_editor/yaml/yaml_parser.py
```diff
@@ -158,6 +162,21 @@ def parse_waveform(self, yaml_str):
...
+                    and isinstance(waveform[0], dict)
+                    and isinstance(waveform[0].get("user_value"), str)
+                ):
+                    static = Waveform(
+                        yaml_str=yaml_str,
+                        line_number=line_number,
+                        name=name,
+                        dd_version=dd_version,
+                    )
+                    static.static_value = waveform[0]["user_value"]
```
SBlokhuizen: This feels a bit hacky, do we have to abuse the waveform class for this?

waveform_editor/tendencies/reference.py
```diff
@@ -0,0 +1,38 @@
...
+    """Tendency that takes its values from an external reference data entry.
+
+    Instead of an analytic shape, the waveform's values are read from a named entry in
+    ``globals.references`` and resampled onto the export time base. By default the same
+    DD path the tendency sits at is read; ``path`` overrides it and ``time_offset``
+    shifts the sampling time. The read/resample is done by the exporter (needs IMAS and
+    the export times), so this class only carries the configuration; it must be the sole
+    tendency of its waveform. Works for scalar and profile (array-per-time) quantities.
```
mikesndrs: In the tests we have a hybrid waveform of linear and reference so this 'must be sole tendency' rule no longer holds right?

waveform_editor/tendencies/reference.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
```
mikesndrs: should be added to documentation

tests/test_exporter.py
```diff
@@ -709,6 +709,144 @@ def test_example_yaml(tmp_path):
...
+    times = np.array([0.0, 1.0])
+    cs = ConfigurationExporter(config, times).to_ids_dict()["core_sources"]
+
+    # sampled at t+1 (CLOSEST): export t=0 -> ext t=1 (10), export t=1 -> ext t=2 (20)
+    assert np.allclose(
+        [cs.source[0].profiles_1d[t].electrons.energy[0] for t in range(2)],
+        [10.0, 20.0],
+    )
+
+
```
mikesndrs, line 1025: might be useful to have a test for expected behavior when referenced IDS does not exist or is empty?

waveform_editor/tendencies/reference.py
DaanVanVugt: It is only mandatory to be the sole tendency if it is a non-0D tendency

waveform_editor/waveform.py
DaanVanVugt: also will extract into separate PR

waveform_editor/configuration.py
DaanVanVugt: yes, and actually we should unify with machine_description and base_idss

tests/test_exporter.py
DaanVanVugt: closest_interp for now, but lets also look at supporting linear

tests/test_exporter.py
DaanVanVugt: ooh, fancy. tricky if the array dimensions change with time, but maybe we can just error in that case

mikesndrs (approved):

DaanVanVugt:
LGTM

waveform_editor/waveform.py
```diff
@@ -29,6 +30,9 @@
     "smooth": SmoothTendency,
     "piecewise": PiecewiseLinearTendency,
     "repeat": RepeatTendency,
+    "import": ImportTendency,
+    # `reference` kept as an alias for the import tendency type.
```
maarten-ic, line 34: Why though? Both are introduced with this PR, so unnecessary for backwards compatibility?

waveform_editor/static_waveform.py
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+
+from waveform_editor.base_waveform import BaseWaveform
+
+
+class StaticWaveform(BaseWaveform):
```
maarten-ic, line 6: This sounds like a contradictio in terminis

waveform_editor/static_waveform.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, value, *, yaml_str="", name="waveform", dd_version=None):
+        super().__init__(yaml_str, name, dd_version)
+        self.yaml_str = yaml_str
+        self.value = value
+
+    def get_value(
+        self, time: np.ndarray | None = None
+    ) -> tuple[np.ndarray, np.ndarray]:
+        # A static value is not a time series; nothing to plot.
```
maarten-ic, line 22: Why not plot a horizontal line at the static value?

maarten-ic: Might be useful when combining with other tendencies

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+INTERP_MODES = ("closest", "linear", "previous")
+
+
+def _interp_const(mode):
+    """Map a user interpolation mode name to an IMAS interpolation constant."""
+    return {
+        "closest": imas.ids_defs.CLOSEST_INTERP,
+        "linear": imas.ids_defs.LINEAR_INTERP,
+        "previous": imas.ids_defs.PREVIOUS_INTERP,
+    }[mode]
```
maarten-ic, line 24: Would be more elegant (and slightly faster) like this:

```python
INTERP_MODES = {
        "closest": imas.ids_defs.CLOSEST_INTERP,
        "linear": imas.ids_defs.LINEAR_INTERP,
        "previous": imas.ids_defs.PREVIOUS_INTERP,
}
```

Then instead of using the `_interp_const(mode)` function, just use `INTERP_MODES[mode]`.

Pros: 
- Maintain single source of modes
- Avoid constructing a new temporary dictionary with every call to `_interp_const`

Cons:
- Need to change merged code

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+    def _source(self, ref):
+        """The import spec (URI string or ``{port: ...}``) for import name ``ref``."""
+        if ref not in self.imports:
+            raise KeyError(f"unknown import '{ref}'")
+        return self.imports[ref]
+
+    @staticmethod
+    def _port_of(source):
+        """The MUSCLE3 port name if ``source`` is a port-import, else None. A
+        port-import is ``{port: <name>}`` or the ``port:<name>`` string shorthand."""
```
maarten-ic, line 60: Would be better to unify when reading the configuration?

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        (loaded into an in-memory entry so it can be sliced like any other)."""
+        dd = self.dd_version
+        port = self._port_of(source)
+        if port is not None:
+            ids = self.received_idss.get(port)
+            if ids is None:
+                raise KeyError(f"no IDS received on import port '{port}'")
+            with imas.DBEntry("imas:memory?path=/", "w", dd_version=dd) as mem:
+                mem.put(imas.convert_ids(ids, dd))
+                yield mem
```
maarten-ic, line 84: This is only used in `_full`, and then the IDS is immediately "get" again? Seems a bit overcomplicated and potential source of bugs?

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        string, or a port-import referring to an IDS received over a MUSCLE3 port
+        (loaded into an in-memory entry so it can be sliced like any other)."""
+        dd = self.dd_version
+        port = self._port_of(source)
+        if port is not None:
+            ids = self.received_idss.get(port)
+            if ids is None:
+                raise KeyError(f"no IDS received on import port '{port}'")
+            with imas.DBEntry("imas:memory?path=/", "w", dd_version=dd) as mem:
+                mem.put(imas.convert_ids(ids, dd))
```
maarten-ic, line 83: Any reason for using explicit conversion for muscle3 IDSs and implicit conversion for data entry sources?

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        if not self._is_homogeneous(full):
+            return full
+        dd = self.dd_version
+        interp_const = _interp_const(interp)
+        with imas.DBEntry("imas:memory?path=/src", "w", dd_version=dd) as src:
+            src.put(full)
+            slices = [
+                src.get_slice(ids_name, float(t) + time_offset, interp_const)
+                for t in time
+            ]
```
maarten-ic, line 120: Better use get_sample instead, that's way more efficient

N.B. this is probably fine for smaller IDSs, but for larger (on-disk) ones I would just directly call `get_sample` instead of doing a full get(), then a put() (to memory) and only then a get_sample

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        A time-independent source (e.g. a machine description) has no time dimension and
+        is returned as-is. The source is hosted in an in-memory entry so it can be
+        resampled slice-by-slice via get_slice regardless of its backend.
+        """
+        full = self._full(ids_name, source)
+        if not self._is_homogeneous(full):
+            return full
+        dd = self.dd_version
+        interp_const = _interp_const(interp)
+        with imas.DBEntry("imas:memory?path=/src", "w", dd_version=dd) as src:
```
maarten-ic, line 115: N.B. the paths aren't very relevant. All memory backend data is deleted after closing the DBEntry at the end of the with-statement


```suggestion
        with imas.DBEntry("imas:memory?path=/", "w", dd_version=dd) as src:
```

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        """The source's own (time, values) at ``ids_path``, without resampling.
+
+        Used for editing/plotting: the raw samples are shown on the source's native time
+        base (mapped into waveform time by subtracting ``time_offset``). Returns empty
+        arrays for a time-independent source, which has no curve to draw.
+        """
+        source = self._source(ref)
+        ids_name, sub = ids_path.split("/", 1)
+        full = self._full(ids_name, source)
+        times = np.asarray(full.time, dtype=float)
```
maarten-ic, line 138: So we cannot load data from IDSs with inhomogeneous time?

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        for part, index in zip(path.parts, path.indices, strict=True):
+            node = node[part]
+            if index is not None:
+                node = node[index]
+        return node
+
+    @staticmethod
+    def _mirror_leaf(src_root, dst_root, full_path):
+        """Copy one leaf, addressed by its concrete root path (e.g.
+        ``source[0]/profiles_1d[3]/electrons/energy``), from src_root into dst_root,
```
maarten-ic, line 272: Seems overly complicated. Why not simplify to:

```python
dst_root[full_path] = src_root[full_path]  # or copy.deepcopy(src_root[full_path]) if needed
```

waveform_editor/import_resolver.py
```diff
@@ -0,0 +1,308 @@
...
+        for name, index in steps[:-1]:
+            src, dst = src[name], dst[name]
+            if index is not None:
+                if len(dst) <= index:
+                    dst.resize(index + 1, keep=True)
+                src, dst = src[index], dst[index]
+        leaf_name = steps[-1][0]
+        dst[leaf_name].value = src[leaf_name].value
+
+    def extract_values(self, node, path, path_index=0):
```
maarten-ic, line 288: Could you reuse the tensorization in e.g. to_xarray? I always get a bit sad when we need to reinvent tensorization

Alternatively we could finalize the IDS Slicing logic in IMAS-Python

docs/source/yaml_format.rst
```diff
@@ -50,21 +50,29 @@ These parameters can be changed under the "Edit Global Properties" tab in the GU
...
+           scenario: imas:hdf5?path=scenario_run
+           live_eq: {port: equilibrium_in}
+
+    Overlay a machine-description IDS with an ``<ids>/*`` wildcard import, then override
+    individual nodes.
+
+    .. code-block:: yaml
+
+       ec_launchers:
+         ec_launchers/*:
```
maarten-ic, line 73: Just to check: we're always taking the default occurrence, right?

docs/source/tendencies.rst
```diff
@@ -188,6 +190,48 @@ Parameters
 .. warning::
     This tendency does **not** accept the common ``start``, ``duration``, or ``end`` parameters. These are derived directly from the required ``time`` list.
 
+.. _import-tendency:
+
+Import
+======
+
+Takes its values from an external entry in :ref:`globals.imports <global_properties>` instead of an analytic shape, resampled onto the export time base. By default it reads the waveform's own DD path.
```
maarten-ic, line 198: What do you mean with "analytic shape"?

maarten-ic:
```suggestion
Takes its values from an external entry in :ref:`globals.imports <global_properties>` instead of an analytic shape, resampled onto the export time base.
```

I think it's more clear to leave this out from the overview text

docs/source/tendencies.rst
```diff
@@ -188,6 +190,48 @@ Parameters
...
+*   ``path``: DD path to read. Defaults to the waveform's own path.
+*   ``time_offset``: Offset added to the export time when sampling. Defaults to ``0``.
+*   ``interp``: Resampling mode: ``closest`` (default), ``linear`` or ``previous``.
+
+.. code-block:: yaml
+
+    core_sources/source(1)/profiles_1d/electrons/energy:
+      - {ref: scenario, interp: linear}
+
+Wildcards expand against the source: a trailing ``*`` imports a whole subtree (``<ids>/*`` copies a whole IDS), and a ``(*)`` index wildcard iterates every element of an array of structure -- several may be combined, e.g. every ion of every source. An overlay may list several sources, applied in order:
```
maarten-ic, line 214: Let's explain this in an example, it took me reading the rest of the PR before I understood what was meant

docs/source/tendencies.rst
```diff
@@ -188,6 +190,48 @@ Parameters
...
+    core_sources/source(*)/profiles_1d/ion(*)/z_ion:
+      - {ref: scenario}
+
+    ec_launchers/*:        # whole-IDS overlay; sources stack
+      - {ref: machine}
+      - {ref: scenario}
+
+**Precedence.** Where imports or explicit waveforms write the same node, the most specific wins regardless of order: ``<ids>/*`` < subtree ``.../*`` < explicit leaf. Equal specificity falls back to listing order (last wins).
+
+Non-0D imports (a profile, a wildcard subtree) own the whole waveform. Only **0D (scalar)** imports combine with analytic segments, each filling its ``[start, end]`` window:
```
maarten-ic, line 227: I don't understand this sentence, trying to unpack it into concrete questions:
1. What is meant with "own the whole waveform"?
2. 0D is a bit confusing. Do you mean truly 0D quantities (such as https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/equilibrium.html#equilibrium-vacuum_toroidal_field-r0) or 0D-over-time?
3. I don't think "analytic segments" is very clear. You just mean you can combine this with other tendencies, as long as there's no wildcards in the waveform name, correct?

docs/source/examples/imports.yaml
```diff
@@ -0,0 +1,49 @@
+# Reference example for the `imports` mechanism. The URIs are illustrative.
+globals:
+  dd_version: 4.0.0
+  imports:
+    machine:  imas:hdf5?path=/path/to/machine_description    # static (no /time)
+    scenario: imas:hdf5?path=/path/to/scenario_run           # time-dependent
```
maarten-ic, line 6: Do you provide these datasets so users can try this for themselves?

maarten-ic (commented):
I had a look at the PR and we should probably do some cleanup post-merge. Could you have a look at the comments and let me know what you think?

docs/source/examples/imports.yaml
```diff
@@ -0,0 +1,49 @@
...
+  # Trailing /* -- mirror every filled leaf of the subtree from the entry.
+  core_sources/source(1)/profiles_1d/*:
+    - {ref: scenario}
+
+  # Index wildcard -- import the leaf for every source.
+  core_sources/source(*)/global_quantities/total_ion_power:
+    - {ref: scenario}
+
+  # Multiple index wildcards -- every (source, ion) combination.
+  core_sources/source(*)/profiles_1d/ion(*)/z_ion:
```
DaanVanVugt, line 35: should index wildcards be `:` instead?

PR #171, opened 2026-06-23
make notification flash duration longer

### LINKED ISSUE #169: error messages disappear too fast
probably not this PR but error messages disappear before I can fully read them

_Originally posted by @mikesndrs in https://github.com/iterorganization/Waveform-Editor/pull/160#discussion_r3453018193_

ioan-alexandra:
fixes #169

### DIFF
--- a/waveform_editor/gui/main.py
+++ b/waveform_editor/gui/main.py
@@ -3,6 +3,7 @@ import logging
 import imas
 import panel as pn
 import param
+from panel.io.notifications import NotificationAreaBase
 
 from waveform_editor.configuration import WaveformConfiguration
 from waveform_editor.gui.dict_editor import DictEditor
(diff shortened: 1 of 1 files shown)


mikesndrs (approved):

waveform_editor/gui/main.py
SBlokhuizen, line 31: Do we want to just extend the error notification duration or all notifications? If the former,  we could just provide a duration here?

waveform_editor/gui/main.py
```diff
@@ -42,6 +43,18 @@ def exception_handler(ex):
     exception_handler=exception_handler,
 )
 
+_NOTIFICATION_DURATION = 20000
```
SBlokhuizen: 20 seconds seems quite long for a notification message to me. By default the notification duration is set to 3 seconds, did you think that was too short? If so, perhaps change it to 5 seconds instead?

mikesndrs: error messages for imas-python backend stuff can be quite long and I had to retrigger the message 4/5 times before I read the full message

waveform_editor/gui/main.py
ioan-alexandra: fairs

SBlokhuizen (approved):

PR #174, opened 2026-06-24
docs update on tendency inference

DaanVanVugt:
When a tendency entry has no explicit `type`, infer it from its keys
instead of always defaulting to linear: `to` -> linear, `time` ->
piecewise, `value` -> constant. Anything else falls back to linear, so a linear tendency does not require `to` (a from-only/rate-only/bare segment is still linear and takes its endpoints from its neighbours).

A segment whose distinguishing key is absent must name its type explicitly: the periodic shapes, smooth, and a value-less constant (a bare {duration} is read as a linear ramp).

Documents the behaviour (yaml_format + tendencies) and tests the inference, the linear-without-`to` forms, the value-less ambiguity, and that an explicit type takes precedence.

### DIFF
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -8,7 +8,7 @@ This document describes the different types of tendencies available in the Wavef
 
 Each tendency defines the behavior of the signal over a specific time interval. You can chain multiple tendencies together to create complex waveforms.
 
-If ``type`` is omitted it is inferred from the entry's keys: ``ref`` → :ref:`import <import-tendency>`, ``to`` → linear, ``time`` → :ref:`piecewise <piecewise-linear-tendency>`, ``value`` → constant; anything else defaults to ``linear`` (so a ``from``-only, ``rate``-only, or bare segment is still a linear ramp). Tendencies with no distinguishing key -- the periodic shapes, ``smooth``, and a value-less ``constant`` -- must name their ``type`` explicitly.
+If ``type`` is omitted it is inferred from the entry's keys: ``ref`` → :ref:`import <import-tendency>`, ``to`` → :ref:`linear <linear-tendency>`, ``time`` → :ref:`piecewise <piecewise-linear-tendency>`, ``value`` → :ref:`constant <constant-tendency>`; anything else defaults to ``linear`` (so a ``from``-only, ``rate``-only, or bare segment is still a linear ramp). Tendencies with no distinguishing key -- the :ref:`periodic shapes <periodic-tendencies>`, :ref:`smooth <smooth-tendency>`, and a value-less :ref:`constant <constant-tendency>` -- must name their ``type`` explicitly.
 
 Common Time Parameters
 ======================
@@ -21,6 +21,8 @@ Most tendencies accept the following parameters to define their time interval. Y
 
 .. note::
     The :ref:`Piecewise Linear Tendency <piecewise-linear-tendency>` is an exception and derives its time interval solely from its ``time`` parameter list. It does *not* accept ``start``, ``duration``, or ``end``.
+
+.. _constant-tendency:
 
 Constant Tendency
 =================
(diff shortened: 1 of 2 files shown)


docs/source/tendencies.rst
```diff
@@ -8,6 +8,10 @@ This document describes the different types of tendencies available in the Wavef
 
 Each tendency defines the behavior of the signal over a specific time interval. You can chain multiple tendencies together to create complex waveforms.
 
+The ``type`` may be omitted, in which case it is inferred from the entry's keys: ``to`` implies a :ref:`Linear <available-tendencies>` tendency, ``time`` a :ref:`Piecewise Linear <piecewise-linear-tendency>` tendency, and ``value`` a :ref:`Constant <available-tendencies>` tendency. Anything else falls back to ``linear`` -- so a linear tendency does *not* require ``to``; a ``from``-only, ``rate``-only, or bare segment is still linear and takes its endpoints from its neighbours.
+
+A segment whose distinguishing key is absent must therefore name its ``type`` explicitly: the periodic shapes (sine, sawtooth, triangle, square), a :ref:`Smooth <available-tendencies>` tendency, and a value-less :ref:`Constant <available-tendencies>` (a bare ``{duration: ...}`` is read as a linear ramp, not a held constant).
+
```
SBlokhuizen: Would be nice to refer to the specific sections in the available-tendencies docpage.

docs/source/yaml_format.rst
```diff
@@ -106,6 +106,8 @@ a list of waveforms, or a single number (float or integer).
                # Implicit linear ramp back to 0 over 25 seconds
              - { duration: 25, to: 0 }
 
+        The ``type`` may be omitted, in which case it is inferred from the entry's keys: ``to`` implies ``linear``, ``time`` implies ``piecewise``, and ``value`` implies ``constant``. Anything else falls back to ``linear`` -- a linear tendency does not require ``to`` (a ``from``-only, ``rate``-only, or bare segment is still linear, taking its endpoints from its neighbours). A segment whose distinguishing key is absent must name its ``type`` explicitly: the periodic shapes, ``smooth``, and a value-less ``constant`` (a bare ``{duration: ...}`` is read as a linear ramp).
+
```
SBlokhuizen: Perhaps drop this and refer to the section `tendencies.rst` instead? Now the same information is written out twice.

SBlokhuizen (commented):
Looks good! I just have 2 minor comments about the documentation.

DaanVanVugt:
later prs introduce new tendencies, we should remember to include new auto-identification for these

mikesndrs (approved):

docs/source/tendencies.rst
```diff
@@ -8,7 +8,7 @@ This document describes the different types of tendencies available in the Wavef
 
 Each tendency defines the behavior of the signal over a specific time interval. You can chain multiple tendencies together to create complex waveforms.
 
-If ``type`` is omitted it is inferred from the entry's keys: ``ref`` → :ref:`import <import-tendency>`, ``to`` → linear, ``time`` → :ref:`piecewise <piecewise-linear-tendency>`, ``value`` → constant; anything else defaults to ``linear`` (so a ``from``-only, ``rate``-only, or bare segment is still a linear ramp). Tendencies with no distinguishing key -- the periodic shapes, ``smooth``, and a value-less ``constant`` -- must name their ``type`` explicitly.
+If ``type`` is omitted it is inferred from the entry's keys: ``ref`` → :ref:`import <import-tendency>`, ``to`` → :ref:`linear <linear-tendency>`, ``time`` → :ref:`piecewise <piecewise-linear-tendency>`, ``value`` → :ref:`constant <constant-tendency>`; anything else defaults to ``linear`` (so a ``from``-only, ``rate``-only, or bare segment is still a linear ramp). Tendencies with no distinguishing key -- the :ref:`periodic shapes <periodic-tendencies>`, :ref:`smooth <smooth-tendency>`, and a value-less :ref:`constant <constant-tendency>` -- must name their ``type`` explicitly.
```
SBlokhuizen, line 11: This would make a lot more sense with a small example in my opinion.

PR #176, opened 2026-06-25
Steps tendency

DaanVanVugt:
Includes a commit from #175

PR #177, opened 2026-06-25
Expressions unification

DaanVanVugt:
After #176


waveform_editor/tendencies/expression.py
```diff
@@ -0,0 +1,135 @@
+import ast
```
SBlokhuizen, line 1: Update docs?

waveform_editor/gui/editor.py
SBlokhuizen, line 91: Don't think this will work anymore with the new expressions.

waveform_editor/yaml/yaml_globals.py
```diff
@@ -6,8 +6,17 @@
 
 logger = logging.getLogger(__name__)
 
+# Configuration schema version. Bumped to 2 when bare-string values became literal
+# string constants instead of derived-waveform expressions.
+CURRENT_SCHEMA_VERSION = 2
+
```
SBlokhuizen, line 12: If we want to have YAML schema versions, should we use semantic versioning instead?

tests/test_yaml/example.yaml
```diff
@@ -8,38 +9,55 @@ dummy_waveform:
   - {to: 0, duration: 100}
 H&CD:
   Powers:
-    ec_launchers/beam(1)/power_launched/data: '"w/1"'
-    ic_antennas/antenna(1)/power_launched/data: '"w/1"'
-    nbi/unit(1)/power_launched/data: |
-      # limits for H or He:
-      16.5e6 * ("nbi/unit(1)/energy/data" / 870e3) ** 2.5
+    ec_launchers/beam(1)/power_launched/data:
+    - expression: '"w/1"'
```
DaanVanVugt: we could also consider a 'raw' value as an expression, but evaluate statically if it is constant (not referring to raw values, not calling external code), and if so turn it into a constant tendency otherwise into an expression tendency?

PR #182, opened 2026-07-23
Type aware constants

SBlokhuizen:
Constant tendencies can now be of different types, namely `int`, `str`, `bool`, and `float`. A waveform will automatically inherit the type of its tendencies, and mixing of tendency types in a single waveform is not allowed (except for `int`s and `float`s in which case the waveform will be considered of type `float`).  When a waveform targets a DD path, its type will be validated against the DD type.

Example of a waveform containing strings: 
image

### DIFF
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -51,6 +51,22 @@ If the ``value`` is not specified, it will be set to the last value of the previ
     - {type: linear, to: 3, duration: 10}
     - {type: constant, duration: 10}
 
+.. _constant-value-types:
+
+Value Types
+-----------
+
+The ``value`` of a constant tendency may be a number, a string, or a boolean:
+
+.. code-block:: yaml
+
+    - {type: constant, value: ohmic, duration: 2}
+    - {type: constant, value: nbi, duration: 2}
+
+.. warning::
+    Integers and floats may be freely combined within a single waveform, but other
+    value types may not be combined with each other.
+
 Linear Tendency
 ===============
 
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -78,7 +78,7 @@ A string value:
     - {type: constant, value: nbi, duration: 2}
 
 It is also possible to use ``True`` and ``False`` as the ``value``, in which case the 
-waveform will be as an integer type (where ``True = 1``, and ``False = 0``):
+waveform will be considered an integer type (where ``True = 1``, and ``False = 0``):
 
 .. code-block:: yaml
 
--- a/docs/source/tendencies.rst
+++ b/docs/source/tendencies.rst
@@ -56,16 +56,39 @@ If the ``value`` is not specified, it will be set to the last value of the previ
 Value Types
 -----------
 
-The ``value`` of a constant tendency may be a number, a string, or a boolean:
+The ``value`` of a constant tendency may be a number (integer or floating-point) or a string.
+
+An integer value:
+
+.. code-block:: yaml
+
+    - {type: constant, value: 3, duration: 2}
+
+A floating-point value:
+
+.. code-block:: yaml
+
+    - {type: constant, value: 3.5, duration: 2}
+
+A string value:
 
 .. code-block:: yaml
 
     - {type: constant, value: ohmic, duration: 2}
     - {type: constant, value: nbi, duration: 2}
 
+It is also possible to use ``True`` and ``False`` as the ``value``, in which case the 
+waveform will be as an integer type (where ``True = 1``, and ``False = 0``):
+
+.. code-block:: yaml
+
+    - {type: constant, value: True, duration: 2}
+    - {type: constant, value: False, duration: 2}
+
 .. warning::
-    Integers and floats may be freely combined within a single waveform, but other
-    value types may not be combined with each other.
+    Integers and floats may be freely combined within a single waveform (the
+    resulting waveform will be float-typed), but it's not allowed to combine string 
+    values with numbers in a waveform.
 
 Linear Tendency
 ===============
--- a/tests/test_dependency_graph.py
+++ b/tests/test_dependency_graph.py
@@ -87,3 +87,22 @@ def test_detect_cycles_with_start_node():
     dg.graph["C"] = {"A"}
     with pytest.raises(RuntimeError):
         dg.detect_cycles("A")
+
+
+def test_topological_order():
+    dg = DependencyGraph()
+    dg.add_node("A", ["B"])
+    dg.add_node("B", ["C"])
+    dg.add_node("C", [])
+
+    order = dg.topological_order()
+    assert set(order) == {"A", "B", "C"}
+    assert order.index("C") < order.index("B") < order.index("A")
+
+
+def test_topological_order_ignores_leaf_dependencies():
+    """Dependencies that are not nodes themselves should not show up."""
+    dg = DependencyGraph()
+    dg.add_node("A", ["leaf"])
+
+    assert dg.topological_order() == ["A"]
--- a/tests/test_dependency_graph.py
+++ b/tests/test_dependency_graph.py
@@ -106,3 +106,17 @@ def test_topological_order_ignores_leaf_dependencies():
     dg.add_node("A", ["leaf"])
 
     assert dg.topological_order() == ["A"]
+
+
+def test_get_dependents():
+    """Test if direct and dependent waveforms are found."""
+    dg = DependencyGraph()
+    dg.add_node("A", [])
+    dg.add_node("B", ["A"])
+    dg.add_node("C", ["B"])
+    dg.add_node("D", [])
+
+    assert dg.get_dependents("A") == ["B", "C"]
+    assert dg.get_dependents("B") == ["C"]
+    assert dg.get_dependents("C") == []
+    assert dg.get_dependents("D") == []
--- a/tests/test_dependency_graph.py
+++ b/tests/test_dependency_graph.py
@@ -95,9 +95,7 @@ def test_topological_order():
     dg.add_node("B", ["C"])
     dg.add_node("C", [])
 
-    order = dg.topological_order()
-    assert set(order) == {"A", "B", "C"}
-    assert order.index("C") < order.index("B") < order.index("A")
+    assert dg.topological_order() == ["C", "B", "A"]
 
 
 def test_topological_order_ignores_leaf_dependencies():
@@ -106,17 +104,3 @@ def test_topological_order_ignores_leaf_dependencies():
     dg.add_node("A", ["leaf"])
 
     assert dg.topological_order() == ["A"]
-
-
-def test_get_dependents():
-    """Test if direct and dependent waveforms are found."""
-    dg = DependencyGraph()
-    dg.add_node("A", [])
-    dg.add_node("B", ["A"])
-    dg.add_node("C", ["B"])
-    dg.add_node("D", [])
-
-    assert dg.get_dependents("A") == ["B", "C"]
-    assert dg.get_dependents("B") == ["C"]
-    assert dg.get_dependents("C") == []
-    assert dg.get_dependents("D") == []
--- a/tests/test_derived_waveform.py
+++ b/tests/test_derived_waveform.py
@@ -209,6 +209,21 @@ def test_derived_waveform_type_mixing():
     assert config["derived_waveform"].annotations  # not allowed to mix str and int
 
 
+def test_derived_waveform_disallows_string_dependency():
+    yaml_str = """
+    root_group:
+      wf1:
+        - {type: constant, value: ohmic, duration: 2}
+      derived_waveform: |
+        'wf1'
+    """
+    config = WaveformConfiguration()
+    config.load_yaml(yaml_str)
+
+    assert config["wf1"].value_type is str
+    assert config["derived_waveform"].annotations
+
+
 def test_derived_waveform_int_float_mixing():
     yaml_str = """
     root_group:
--- a/tests/test_derived_waveform.py
+++ b/tests/test_derived_waveform.py
@@ -154,21 +154,21 @@ def test_function_access_control(filled_config):
                 waveform.get_value(time_ret)
 
 
-def test_derived_waveform_type_matches_original(config):
-    original_name = "wf1"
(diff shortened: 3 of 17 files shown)


waveform_editor/tendencies/constant.py
```diff
@@ -65,6 +66,11 @@ def _calc_values(self):
         else:
             value = self.user_value
 
+        # If value is inherited from previous tendency, value normalize it back to
+        # a plain Python type
+        if isinstance(value, np.generic):
+            value = value.item()
```
maarten-ic: Two options:
1. Move immediately after line 65 (`value = self.prev_tendency.end_value`) to keep the logic together
2. Update `Tendency.end_value` so it always returns a python type instead of a numpy scalar

waveform_editor/tendencies/repeat.py
```diff
@@ -28,7 +28,20 @@ def __init__(self, **kwargs):
 
         self.waveform = Waveform(waveform=waveform, is_repeated=True)
         self.period = 1
+        # Categorical values are not supported inside a repeat tendency.
+        has_categorical_value = any(
+            t.value_type in (str, bool) for t in self.waveform.tendencies
+        )
+        if has_categorical_value:
+            self.waveform.tendencies = []
         super().__init__(**kwargs)
```
maarten-ic, line 31: Any reason that the has_categorical_value is split by this line? I suppose `super().__init__()` should happen before the error check, but can we move it just after `self.period` and only have a single `if self.waveform.value_type in (str, bool)`?

waveform_editor/base_waveform.py
```diff
@@ -9,6 +9,8 @@
 
 
 class BaseWaveform(ABC):
+    value_type = float
```
maarten-ic: Why is this not an instance member?

waveform_editor/derived_waveform.py
```diff
@@ -78,6 +78,7 @@ def prepare_expression(self):
         """Parse the YAML expression, extract dependencies, transform it for
         evaluation, and compile it.
         """
+        self.annotations.clear()
```
maarten-ic, line 83: Explain what this is needed for?

waveform_editor/derived_waveform.py
```diff
@@ -110,6 +112,35 @@ def rename_dependency(self, old_name, new_name):
...
+        if missing:
+            self.annotations.add(0, f"Unknown dependency: {sorted(missing)!r}\n")
+            return
+
+        if len(dependency_types) > 1:
+            type_names = sorted(t.__name__ for t in dependency_types)
+            self.annotations.add(
+                0,
+                "All dependencies of a derived waveform must have the same "
+                f"type. Found: {type_names}\n",
```
maarten-ic: I don't agree with this: mixing ints and floats should just work (e.g. 1 * 2.3 = 4.6).
Also I could imagine that boolean waveforms are useful in conditional expressions: `waveform1 if boolean_waveform else waveform2`

waveform_editor/derived_waveform.py
```diff
@@ -110,6 +112,35 @@ def rename_dependency(self, old_name, new_name):
         self.yaml = renamer.yaml
         self.prepare_expression()
 
+    def _validate_type(self):
```
maarten-ic: This validation now depends on properties of other waveforms (rather than just existence). Is this re-evaluated when the type of a dependent waveform changes?

waveform_editor/derived_waveform.py
```diff
@@ -110,6 +112,35 @@ def rename_dependency(self, old_name, new_name):
...
+
+        dependency_types = set()
+        missing = set()
+        for dependency in self.dependencies:
+            try:
+                dependency_types.add(self.config[dependency].value_type)
+            except KeyError:
+                missing.add(dependency)
+
+        if missing:
```
maarten-ic, line 133: This is also checked elsewhere, isn't it? Let's make sure we don't duplicate the logic

waveform_editor/waveform.py
```diff
@@ -15,6 +16,21 @@
 from waveform_editor.tendencies.repeat import RepeatTendency
 from waveform_editor.tendencies.smooth import SmoothTendency
 
+IDS_DATATYPE_MAP = {
+    float: IDSDataType.FLT,
+    str: IDSDataType.STR,
+    int: IDSDataType.INT,
+    bool: IDSDataType.INT,  # Booleans don't exist in DD
```
maarten-ic: Why don't we follow suite and use integer waveforms also for boolean expressions? What is the benefit of having boolean expressions?

Benefit of using integers is that plotting "just works" and displays a 1 or a 0

waveform_editor/waveform.py
```diff
@@ -173,11 +196,51 @@ def _process_waveform(self, waveform):
...
+        if not self.tendencies:
+            return
+
+        self.value_type = self.tendencies[0].value_type
+        for tendency in self.tendencies[1:]:
+            if {tendency.value_type, self.value_type} <= {int, float}:
+                if tendency.value_type is float:
+                    self.value_type = float
+                continue
+            if tendency.value_type != self.value_type:
```
maarten-ic: This is a bit cryptic why not:

```python
value_types = set(tendency.value_type for tendency in self.tendencies)
if len(value_types) == 1:
    self.value_type = value_types.pop()
elif value_types == {int, float}:
    self.value_type = float
else:
    error with all value types used in the waveform
```

waveform_editor/waveform.py
```diff
@@ -173,11 +196,51 @@ def _process_waveform(self, waveform):
...
+            return
+
+        # An int value is also valid for a float field
+        int_for_flt = (
+            self.value_type is int and self.metadata.data_type is IDSDataType.FLT
+        )
+        if (
+            not int_for_flt
+            and IDS_DATATYPE_MAP[self.value_type] != self.metadata.data_type
+        ):
```
maarten-ic: Let's add IDSDataType.FLT as an allowed datatype for ints in the map and simplify this logic?

```python
if self.metadata is not None and self.metadata.data_type not in IDS_DATATYPES[self.value_type]:
    error
```

maarten-ic (commented):
Couple of comments, I didn't look at the tests yet. Let me know if you want to discuss anything!

waveform_editor/tendencies/base.py
```diff
@@ -227,7 +227,11 @@ def _get_value_and_derivative(self, time):
         """Get the value and derivative of the tendency at a given time."""
         _, value_array = self.get_value(np.array([time]))
         derivative_array = self.get_derivative(np.array([time]))
-        return value_array[0], derivative_array[0]
+        value = value_array[0]
+        # Normalize numpy scalars back to plain Python types
+        if isinstance(value, np.generic):
+            value = value.item()
+        return value, derivative_array[0]
```
maarten-ic: if `value_array` and `derivative_array` are guaranteed numpy arrays, then we could just do:
```suggestion
        # Normalize numpy scalars back to plain Python types
        return value_array.item(0), derivative_array.item(0)
```

Do we have cases where `get_value()` and `get_derivative()` return something else?

waveform_editor/dependency_graph.py
```diff
@@ -101,6 +101,25 @@ def rename_node(self, old_name, new_name):
...
+        """
+        dependents = set()
+        stack = [name]
+        while stack:
+            current = stack.pop()
+            for node, deps in self.graph.items():
+                if current in deps and node not in dependents:
+                    dependents.add(node)
+                    stack.append(node)
+        return [node for node in self.topological_order() if node in dependents]
```
maarten-ic: Do we need this complexity, or could we just traverse all nodes in topological order and do another prepare_expression?

waveform_editor/derived_waveform.py
```diff
@@ -114,7 +116,8 @@ def rename_dependency(self, old_name, new_name):
 
     def _validate_type(self):
```
maarten-ic: Name should reflect that this checks for missing dependencies also

waveform_editor/derived_waveform.py
```diff
@@ -131,15 +134,17 @@ def _validate_type(self):
...
+        elif len(dependency_types) == 1:
+            self.value_type = dependency_types.pop()
+        else:
             type_names = sorted(t.__name__ for t in dependency_types)
             self.annotations.add(
                 0,
                 "All dependencies of a derived waveform must have the same "
-                f"type. Found: {type_names}\n",
+                f"type, or be a mix of int and float. Found: {type_names}\n",
             )
```
maarten-ic: So, basically, you cannot use string waveforms as a dependency? Then let's just implement that and make the error more clear?

Or do we have a use case for DerivedWaveform with string waveforms as input and string waveforms as output?

docs/source/tendencies.rst
```diff
@@ -51,6 +51,22 @@ If the ``value`` is not specified, it will be set to the last value of the previ
     - {type: linear, to: 3, duration: 10}
     - {type: constant, duration: 10}
 
+.. _constant-value-types:
+
+Value Types
+-----------
+
+The ``value`` of a constant tendency may be a number, a string, or a boolean:
```
maarten-ic: Is a `value=True` converted to `value=1`? (and False to 0), or should users not use True/False values?

docs/source/tendencies.rst
```diff
@@ -51,6 +51,22 @@ If the ``value`` is not specified, it will be set to the last value of the previ
...
+
+.. code-block:: yaml
```
maarten-ic, line 75: Show number (float/int) in example as well?

docs/source/tendencies.rst
```diff
@@ -51,6 +51,22 @@ If the ``value`` is not specified, it will be set to the last value of the previ
...
+
+    - {type: constant, value: ohmic, duration: 2}
+    - {type: constant, value: nbi, duration: 2}
+
+.. warning::
+    Integers and floats may be freely combined within a single waveform, but other
+    value types may not be combined with each other.
```
maarten-ic:
```suggestion
    Integers and floats may be freely combined within a single waveform, but it's not
    allowed to combine string values with numbers in a waveform.
```

tests/test_dependency_graph.py
```diff
@@ -87,3 +87,36 @@ def test_detect_cycles_with_start_node():
...
+
+def test_topological_order():
+    dg = DependencyGraph()
+    dg.add_node("A", ["B"])
+    dg.add_node("B", ["C"])
+    dg.add_node("C", [])
+
+    order = dg.topological_order()
+    assert set(order) == {"A", "B", "C"}
+    assert order.index("C") < order.index("B") < order.index("A")
```
maarten-ic: order is a list, right? This graph only has a single topological order:

```python
assert dg.topological_order() == ["C", "B", "A"]
```

tests/test_dependency_graph.py
```diff
@@ -87,3 +87,36 @@ def test_detect_cycles_with_start_node():
...
+
+
+def test_topological_order_ignores_leaf_dependencies():
+    """Dependencies that are not nodes themselves should not show up."""
```
maarten-ic, line 102: Put in docstring of Graph.topological_order() as well

tests/test_derived_waveform.py
```diff
@@ -151,3 +152,98 @@ def test_function_access_control(filled_config):
...
+    wf2 = config["wf2"]
+    wf1 = config["wf1"]
+    assert wf4.value_type is str
+    assert wf3.value_type is str
+    assert wf2.value_type is str
+    assert wf1.value_type is str
+    assert not wf4.annotations
+    assert not wf3.annotations
+    assert not wf2.annotations
+    assert not wf1.annotations
```
maarten-ic: DRY?

```suggestion
    for wf in ["wf1", "wf2", "wf3", "wf4"]:
        assert config[wf].value_type is str
        assert not config[wf].annotations
```

tests/test_derived_waveform.py
```diff
@@ -151,3 +152,98 @@ def test_function_access_control(filled_config):
...
+
+
+def test_derived_waveform_type_mixing(config):
```
maarten-ic: I like the style of the previous test better (just define a config with all waveforms), would that work here as well?

tests/test_derived_waveform.py
```diff
@@ -151,3 +152,98 @@ def test_function_access_control(filled_config):
...
+    assert wf2.value_type is str
+    config.add_waveform(wf2, ["root_group"])
+
+    derived_name = "derived_waveform"
+    yaml_str = f"{derived_name}: |\n  '{wf1_name}' + '{wf2_name}'"
+    derived = DerivedWaveform(yaml_str, derived_name, config)
+    assert derived.annotations  # not allowed to mix str and int type waveforms
+
+
+def test_derived_waveform_int_float_mixing(config):
```
maarten-ic: Idem

tests/test_exporter.py
```diff
@@ -647,6 +647,43 @@ def test_export_constant(tmp_path):
...
+      - {type: constant, value: 1, duration: 2}
+      core_profiles/profiles_1d/grid/psi_magnetic_axis:
+      - {type: constant, value: 1.5, duration: 2}
+      - {type: constant, value: 3.0, duration: 2}
+      core_profiles/profiles_1d/ion(1)/name:
+      - {type: constant, value: D, duration: 2}
+      - {type: constant, value: He, duration: 2}
+      core_profiles/profiles_1d/ion(1)/multiple_states_flag:
+      - {type: constant, value: 1, duration: 2}
+      - {type: constant, value: 0, duration: 2}
```
maarten-ic, line 668: Okay for testing, but functionally that makes no sense 😉 

The only time-dependent STR that I could find in DD 4.1.1 was https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ece.html#ece-channel-beam_tracing-beam-mode

(Makes me wonder why we need it at all, but let's proceed anyway)

maarten-ic (commented):
Couple more comments, let me know if you want to discuss any!

docs/source/tendencies.rst
SBlokhuizen: You can indeed use boolean values, and they will internally be treated integers. I have added an example to the docs to be more clear on this.

maarten-ic (approved):

PR #183, opened 2026-07-31
Infer tendency type by keys

SBlokhuizen:
Infer a tendency's type based on what keys are present in the tendency.

### DIFF
--- a/waveform_editor/waveform.py
+++ b/waveform_editor/waveform.py
@@ -45,6 +45,33 @@ tendency_map = {
     "repeat": RepeatTendency,
 }
 
+INFERRED_TYPE_BY_KEY = {
+    "user_time": "piecewise",
+    "user_value": "constant",
+    "user_waveform": "repeat",
+    "user_base": "sine",
+    "user_amplitude": "sine",
+    "user_min": "sine",
+    "user_max": "sine",
+    "user_phase": "sine",
+}
+
+
+def _infer_tendency_type(entry):
+    """Infer a tendency's type from keys in the tendency entry, defaulting to
+    linear if nothing distinctive is present.
+
+    Args:
+        entry: Entry in the YAML file.
+
+    Returns:
+        The inferred tendency type.
+    """
+    for key, tendency_type in INFERRED_TYPE_BY_KEY.items():
+        if key in entry:
+            return tendency_type
+    return "linear"
+
 
 class Waveform(BaseWaveform):
     def __init__(
@@ -252,10 +279,6 @@ class Waveform(BaseWaveform):
         line_number = entry.get("line_number", 0)
         ignore_msg = "This tendency will be ignored.\n"
 
-        # If no type is given, take linear as default
-        if "user_type" not in entry:
-            entry["user_type"] = "linear"
-
         tendency_type = entry.get("user_type", None)
         if tendency_type is None:
             error_msg = f"The tendency type cannot be empty.\n{ignore_msg}"
@@ -305,6 +328,10 @@ class Waveform(BaseWaveform):
         Returns:
             The created tendency or None, if the tendency cannot be created
         """
+        # If no type is given, infer it from the entry's keys
+        if "user_type" not in entry:
+            entry["user_type"] = _infer_tendency_type(entry)
+
         if self._has_type_error(entry):
             return None
         else:
--- a/waveform_editor/waveform.py
+++ b/waveform_editor/waveform.py
@@ -302,13 +302,9 @@ class Waveform(BaseWaveform):
                 ignore_msg = "This tendency will be ignored.\n"
                 if user_type is None:
                     error_msg = f"The tendency type cannot be empty.\n{ignore_msg}"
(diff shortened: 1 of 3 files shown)


waveform_editor/waveform.py
```diff
@@ -45,6 +45,33 @@
...
 
+INFERRED_TYPE_BY_KEY = {
+    "user_time": PiecewiseLinearTendency,
+    "user_value": ConstantTendency,
+    "user_waveform": RepeatTendency,
+    "user_base": SineWaveTendency,
+    "user_amplitude": SineWaveTendency,
+    "user_min": SineWaveTendency,
+    "user_max": SineWaveTendency,
+    "user_phase": SineWaveTendency,
```
maarten-ic: Is the sine wave a more sensible default than the square or triangle waves? If we don't have practical examples let's maybe not define them and add later based on perceived usage?

waveform_editor/waveform.py
```diff
@@ -305,10 +328,14 @@ def _handle_tendency(self, entry):
...
+        # If no type is given, infer it from the entry's keys
+        if "user_type" not in entry:
+            return _infer_tendency_class(entry)(**entry)
+
         if self._has_type_error(entry):
             return None
         else:
             tendency_type = entry.pop("user_type")
-            tendency_class = tendency_map[tendency_type]
+            tendency_class = TENDENCY_MAP[tendency_type]
```
maarten-ic: Maybe turn this around?

e.g.
```python
if "user_type" not in entry:
    tendency_class = _infer_tendency_class(entry)
else:
    tendency_class = TENDENCY_MAP.get(entry.pop("user_type"))
    if tendency_class is None:
        # error handling here, maybe inline or refactor _has_type_error -> _handle_type_error
        return None
return tendency_class(**entry)
```

waveform_editor/waveform.py
```diff
@@ -252,10 +279,6 @@ def _has_type_error(self, entry):
         line_number = entry.get("line_number", 0)
         ignore_msg = "This tendency will be ignored.\n"
 
-        # If no type is given, take linear as default
-        if "user_type" not in entry:
-            entry["user_type"] = "linear"
-
         tendency_type = entry.get("user_type", None)
         if tendency_type is None:
             error_msg = f"The tendency type cannot be empty.\n{ignore_msg}"
```
maarten-ic: So, it's fine to leave out `type` but leaving it blank is an error?

maarten-ic (commented):
Some questions, let me know what you think!

waveform_editor/waveform.py
SBlokhuizen: To me sine wave seems like the most sensible default for the periodic waveforms indeed. However, I don't have a strong opinion on this, so I am fine to remove these defaults.

waveform_editor/waveform.py
SBlokhuizen: Indeed, I don't see a relevant use case for leaving `type` empty, and I would assume this will usually happen if someone forgot to fill in the tendency type.

waveform_editor/waveform.py
```diff
@@ -330,12 +291,30 @@ def _handle_tendency(self, entry):
...
+            )
+            if tendency_class is None:
+                line_number = entry.get("line_number", 0)
+                ignore_msg = "This tendency will be ignored.\n"
+                if user_type is None:
+                    error_msg = f"The tendency type cannot be empty.\n{ignore_msg}"
+                elif not isinstance(user_type, str):
+                    error_msg = (
+                        f"The tendency type should be of type 'string'.\n{ignore_msg}"
+                    )
```
maarten-ic: I actually think that `Unsupported tendency type: '1'. [...]`. is a more useful error than `The tendency type should be of type 'string'` (which contains type twice, meaning something else in both usages)

```suggestion
```

maarten-ic (commented):
Nice, always good to see complexity reduced! 😄

waveform_editor/waveform.py
SBlokhuizen: I have simplified the logic a bit, let me know what you think!

waveform_editor/waveform.py
```diff
@@ -293,28 +293,15 @@ def _handle_tendency(self, entry):
         if "user_type" not in entry:
             tendency_class = _infer_tendency_class(entry)
         else:
-            user_type = entry.pop("user_type")
-            tendency_class = (
-                TENDENCY_MAP.get(user_type) if isinstance(user_type, str) else None
-            )
+            user_type = str(entry.pop("user_type"))
```
maarten-ic: Much simpler, thanks!

Would it be more useful to have an error message:
`Unsupported tendency type: ''` then `Unsupported tendency type: 'None'`? There's no None in the yaml file

```suggestion
            user_type = entry.pop("user_type")
            user_type = "" if user_type is None else str(user_type)
```

maarten-ic (approved):
LGTM!

PR #184, opened 2026-07-31
Steps tendency

SBlokhuizen:
Introduces a new tendency type, the `steps` tendency. This is basically a shorthand way to write a sequence of constant tendencies.


docs/source/tendencies.rst
```diff
@@ -227,6 +227,53 @@ Parameters
...
+writing out a sequence of :ref:`Constant Tendencies <constant-value-types>`.
+
+Parameters
+----------
+*   ``time``: A list of the start times of each step. Must be strictly monotonically increasing and must have at least 1 point.
+*   ``value``: A list of the values held during each step. Must have the same length as ``time``.
+    Just like the :ref:`Constant Tendency <constant-value-types>`, the ``value`` list may contain numbers or strings:
+*   ``duration``, ``end``: See :ref:`Common Time Parameters <available-tendencies>`. 
+    Either may be used to specify when the last step ends. 
+    If both are omitted, the tendency simply stops at the last point in the ``time`` list.
```
maarten-ic, line 246: This is a bit misleading, since we'll do constant extrapolation at the end of the tendency regardless right?

docs/source/tendencies.rst
```diff
@@ -227,6 +227,53 @@ Parameters
...
+.. image:: images/steps_1.png
+   :alt: Example plot of a Steps Tendency
+   :width: 400px
+   :align: center
+
+.. code-block:: yaml
+
+    - {type: steps, time: [0, 2, 4], value: [1, 3, 5], end: 6}
+
+The same tendency can equivalently be written using ``duration`` instead of ``end``:
```
maarten-ic, line 257: But only because the first time value is 0?

waveform_editor/tendencies/piecewise.py
```diff
@@ -3,6 +3,7 @@
 
 from waveform_editor.annotations import Annotations
 from waveform_editor.tendencies.base import BaseTendency
+from waveform_editor.tendencies.util import validate_time_array
 
 
 class PiecewiseLinearTendency(BaseTendency):
```
maarten-ic, line 9: Would it make sense to make a single superclass for this one and the steps tendency?

The main difference is the interpolation strategy between points. Perhaps we want a "closest" interpolation strategy in the future or "pspline" or ...

The superclass could take care of most of the logic, then the concrete subclasses just provide the interpolation and derivative logic.
See maybe: https://docs.scipy.org/doc/scipy/tutorial/interpolate/1D.html#tutorial-interpolate-interp1d-replacements

waveform_editor/tendencies/util.py
```diff
@@ -1,6 +1,72 @@
...
+
+    Returns:
+        The merged type, or None if the types cannot be merged.
+    """
+    types = set(types)
+    if types <= {int, float}:
+        return float if float in types else int
+    if len(types) == 1:
+        return types.pop()
+    return None
```
maarten-ic, line 20: Could we just follow numpy behaviour for stepwise tendencies?

```python
>>> numpy.array([1, 2, 3])
array([1, 2, 3])
>>> numpy.array([1, 2, 3.])
array([1., 2., 3.])
>>> numpy.array([1, 2, 'x'])
array(['1', '2', 'x'], dtype='<U21')
```

obviously you cannot use the last one with linear interpolation, but I'd be fine casting everything to strings if you have a string value

maarten-ic (commented):
Some highlevel comments, let's agree on those first

docs/source/tendencies.rst
```diff
@@ -227,6 +227,43 @@ Parameters
...
+Steps Tendency
+==============
+
+Defines a sequence of steps, each held constant from its time point until the next one.
+``time`` and ``value`` must have the same length. This is a compact
+alternative to writing out a sequence of :ref:`Constant Tendencies <constant-value-types>`.
+
+Parameters
+----------
+*   ``time``: A list of the tendency's time points. Must be strictly monotonically increasing and must have at least 1 point.
```
mikesndrs: when using this tendency as the 3rd in a sequence, are the time values absolute, relative to t0 or relative to the start of this tendency?

docs/source/tendencies.rst
```diff
@@ -227,6 +227,43 @@ Parameters
...
+*   ``value``: A list of the values held during each step. Must have the same length as ``time``.
+
+.. image:: images/steps_1.png
+   :alt: Example plot of a Steps Tendency
+   :width: 400px
+   :align: center
+
+.. code-block:: yaml
+
+    - {type: steps, time: [0, 2, 4, 6], value: [1, 3, 5, 5]}
```
mikesndrs, line 252: wouldn't it be more intuitive to do values: {0: 1, 2: 3, 4: 5, 6: 5}?
This would be super annoying to work with for longer arrays

docs/source/tendencies.rst
SBlokhuizen: Similarly to how the piecewise linear tendency worked, the times are absolute. I have updated the docs to clarify this.

docs/source/tendencies.rst
SBlokhuizen: I have kept the same notation for the steps tendency as for the existing piecewise linear tendency. While the dict approach is a bit more compact, the existing approach is easier to work with if you want to copy and paste large arrays of numbers from elsewhere. If there is the demand, we can always support both cases at a later point.

mikesndrs (approved):

PR #185, opened 2026-08-03
Shorthand notation

SBlokhuizen:
Updates parsing of shorthand string, float and integer notation as a waveform containing a single constant tendency.

### DIFF
--- a/waveform_editor/yaml/yaml_parser.py
+++ b/waveform_editor/yaml/yaml_parser.py
@@ -165,6 +165,26 @@ class YamlParser:
                     name=name,
                     dd_version=dd_version,
                 )
+            # Shorthand notation for waveforms containing a constant tendency, e.g.
+            #  ec_launchers/beam(1)/phase/angle: -1.65898  # float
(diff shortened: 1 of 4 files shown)


waveform_editor/yaml/yaml_parser.py
```diff
@@ -165,6 +165,26 @@ def parse_waveform(self, yaml_str):
...
+                            "user_type": "constant",
+                            "user_value": waveform,
+                            "line_number": line_number,
+                        }
+                    ],
+                    yaml_str=yaml_str,
+                    line_number=line_number,
+                    name=name,
+                    dd_version=dd_version,
+                )
```
maarten-ic, line 187: Is this exported to a waveform with single constant tendency again?

Would it be better to have a `ConstantWaveform` class instead?

maarten-ic: Btw, this used to be handled as a derived waveform without dependencies, right? Can you explain why we need this distinction?

SBlokhuizen:
Closed in favor of the implementation in #191.

PR #186, opened 2026-08-03
Export to static IDS nodes

SBlokhuizen:
Allow exporting to non-time-dependent 0D IDS nodes. These waveforms may only consist of a single constant tendency, e.g.

```yaml
equilibrium/vacuum_toroidal_field/r0:
  - {type: constant, value: 6.2}
equilibrium/code/name:
  - {type: constant, value: "NICE"}
equilibrium/ids_properties/homogeneous_time:
  - {type: constant, value: 0}
```

### DIFF
--- a/tests/test_exporter.py
+++ b/tests/test_exporter.py
@@ -647,6 +647,24 @@ def test_export_constant(tmp_path):
         assert np.array_equal(ids.beam[2].phase.angle, [3.3e3] * 3)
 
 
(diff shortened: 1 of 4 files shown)


maarten-ic:
Could we reduce complexity by introducing a ConstantWaveform subclass (see shorthand notation) and require that instead of checking for a single constant tendency in the export?

SBlokhuizen:
Closed in favor of the implementation in #191.

PR #187, opened 2026-08-20
Add relative imas paths

Yannicked:


SBlokhuizen (approved):

waveform_editor/import_resolver.py
```diff
@@ -52,7 +55,35 @@ def _source(self, ref):
         """The import spec (URI string or ``{port: ...}``) for import name ``ref``."""
         if ref not in self.imports:
             raise KeyError(f"unknown import '{ref}'")
-        return self.imports[ref]
+        return self._resolve_uri(self.imports[ref])
+
+    def _resolve_uri(self, source):
```
maarten-ic, line 60: This feels a bit weird solution to me, could we think of something better?

1. It doesn't work with netCDF files (though maybe that fails regardless with the waveform editor?)
2. It probably fails for the UDA backend
3. But most unexpectedly: your waveforms change (or won't work anymore) in the following scenarios:
  a. Save configuration for the first time (when editing in the GUI)
  b. Save the configuration in a different folder

